PR #1, opened 2019-07-18
Add keywords and switch rv type to <:Sampleable

FriesischScott:


FriesischScott:
The problem with using `<:Sampleable` is that it would allow multivariate distributions which does not make a lot of sense in a set of random variables with possible correlation.

Maybe `<:Sampleable{Univariate}` works?

AnderGray:
Well spotted, and the suggestion works

PR #4, opened 2019-10-28
Create inputs and basic project structure

### LINKED ISSUE #2: Create necessary inputs
Apart from `RandomVariableSet` what other inputs do we need?

- `RandomVariable`: No need to wrap the objects from `Distributions.jl`.
- `Parameter`: This could potentially just be any subtype of `Number` and just be filled into the `DataFrame`
- `Function`: If we do this, it should probably just take a simple `lambda` function that evaluates one variable and adds it to the `DataFrame`. Alternatively this could be omitted and people would need to just include it in their model/solver. Might be nice though for intermediate values or something you might want in your end results.

Should we have a common `type` for everything that can be an input? This type could be sub-typed together with what we need from `Distributions.jl`.

### LINKED ISSUE #3: Solvers and Models
Regarding the overall structure I think what we should do is something like this:

- A `model` connects several `inputs` to one or multiple `solvers?`. 
- The `solvers` are executed in order. The most basic solver being a `julia` function that you pass in. Somewhat like the `Mio` we have in `OpenCossan`.
- Everything just passes around `DataFrames`. A Solver would just run whatever it needs to and adds the results to the `DataFrame` of the inputs. The end result would then collect all inputs and output values for whatever one might want to do with them afterwards.

#### (Pseudo-)Example:
```julia
a = 5
b = Normal()

function add (df::DataFrame) 
    df.c = df.a .+ df.b
    return df
end

solver = Solver(add)

model = Model([a b], ["a" "b"], [solver])

result = run(model, samples=1000)
```

Something like this, only more complex :thinking:.

This could then be extended to include `ObjectiveFunctions` for reliability analysis or advanced sampling algorithms, etc.

FriesischScott:
This is my proposal for a base layout of the project. It adds basic structures for the required inputs and a method to calculate a probability of failure using Monte Carlo simulation.

I decided to wrap `Distributions` in a `RandomVariable` in order to assign names to all inputs. It preserves complete functionality of `Distributions`. `RandomVariables` can be used independently from the `RandomVariableSet`. This way independent variables can be excluded from the set. I also created a `Parameter` which just wraps a `value::Real` and a `name`. Attaching the name to the random variable also simplifies the rvset.

Since `Function` and `MIO` in OpenCossan basically do the same, I decided to drop the `Function` completely. If the user needs to have intermediate results, models can be split and passed as a list to `probabilityOfFailure` (see demo).

The goal is to allow for Julia's multiple dispatch to figure out which way to compute the pf based on which `simulation` structure is passed. 

I updated `demo/cantilever.jl` to show an example script using all the new structures/functions.

Thoughts?

Closes #2 and closes #3


AnderGray:
So great call on removing the MIO function, we really have no need for it. 
And I agree with everything that you have done.

I've made a couple of changes:

- Abstract types: I've created an abstract type called UQtypes, which all of our types will inherit. This is to allow us to work with Julia multiple dispatch functionality. I've created 3 sub abstract types. `AbstractInput`, `AbstractModel` and `AbstractSimulation`. I'm not sure if we're going to need multiple `model` types, but its there just in case.

- The samples function can now take singular inputs. This was done with just a wrapper function.

- Probability of failure can also take singular inputs, as well as the arrays. This was done with a type Union in it's input

- Reeport package has been added so we can auto-export the distributions package

I've also added a function for the `model` type. This allows you to just call the variable as if it were a function. Example:

```Julia
julia> model1 = Model(df -> df.b .* df.h .^ 3 / 12, "I");
julia> df = DataFrame(b = 0.3, h = 0.4);
julia>ou1 = model1(df)
1-element Array{Float64,1}:
 0.0016000000000000005
```

This just does the same as the evaluate function, but I think this is a cool Julia feature

FriesischScott:
I think we should reexport as little modules as possible. I agree that reexporting Distributions makes sense, since its such an integral part the whole UQ thing. However, I don't think we should reexport `DataFrames`, as it is only necessary if you want to do additional manipulation of the `DataFrame` after the simulation is done.

I like being able to call the `Model` like a function, nicely done! Is there any documentation on that `function ()()` double parentheses syntax? I couldn't find any but it looks interesting.

FriesischScott:
What I do not agree with is the `AbstractSimulation` type. Or at least I don't see any use for it (yet). The problem with the way you implemented it now is that it goes against how multiple dispatch works. 

By using `sim::AbstractSimulation`, Julia can not determine at run-time which simulation is required and we would need to `switch` on the type of simulation to distinguish between for example standard Monte Carlo and Line Sampling for computing the probability of failure. The way it should work is that you define multiple methods with different simulation types like so:

```julia
function probabilityOfFailure(
    models::Union{Array{<:AbstractModel}, AbstractModel},
    performance::Function,
    inputs::Union{Array{<:AbstractInput}, AbstractInput},
    sim::MonteCarlo,
)
	# compute pf with standard Monte Carlo
end

function probabilityOfFailure(
    models::Union{Array{<:AbstractModel}, AbstractModel},
    performance::Function,
    inputs::Union{Array{<:AbstractInput}, AbstractInput},
    sim::LineSampling,
)
	# compute pf with Line Sampling
end
```
Then, depending on what `Simulation` object you pass, the correct method is run. Or am I wrong?

It does make perfect sense for the `AbstractInput` and `AbstractModel` types.

FriesischScott:
@AnderGray If you agree with this, I will write some tests and merge to master and we can keep going from a good basis. If you don't, we can of course iterate further.

AnderGray:
@FriesischScott I'm happy to merge this with the master.

Also you're totally right about the abstract type for simulations not making sense. 

I can't remember where I saw it, I think it was on a Julia presentation... and I can't find it anywhere in the documentation anyway.

PR #6, opened 2019-11-18
Implement line sampling for failure probability

FriesischScott:
This adds `LineSampling` and the required methods such as computation of the gradient and mapping to standard normal/physical space. New dependencies are:

- `FiniteDifferences.jl` to calculate gradients
- `Dierckx.jl` to interpolate and find the root on the line when calculating the pf with `LineSampling`

I also introduced two new types for deterministic and random inputs. This allows us to better filter for random/ non-random inputs.

In addition, I switched all names from `::String` to `::Symbol` to make working with the `DataFrames` easier.

I'm not 100% sure about some of the method names. Especially for the mappings between the spaces.


FriesischScott:
I am going to merge this now. @AnderGray if you have any remarks/changes you'd like to see, let me know.

PR #10, opened 2020-04-27
Rewrite RandomVariableSet to JointDistribution

FriesischScott:
Instead of the `RandomVariableSet` this introduces two new types:

- `JointDistribution`
- `Copula`

A `JointDistribution` is built from an array of marginal distributions and a copula of the same dimension. Currently the only implementation of `Copula` is the `GaussianCopula`.


AnderGray:
100% the right move to move from `RandomVariableSet ` to `Copulas` and `JointDistributions`.

The ` to_standard_normal_space` and `to_copula_space` are very clean. I didn't know you could do them like that.

What are you using FiniteDifferences.jl for?

FriesischScott:
`FiniteDifferences` is used to compute the gradients needed for LineSampling. Or if you just want to do local sensitivity.

PR #12, opened 2020-08-05
Switch from Travis to Github Actions

FriesischScott:
Github Actions allows for parallel tests on Linux, Mac and Windows. Plus having it self contained in Github is nice.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+on:
+  - push
+  - pull_request
+jobs:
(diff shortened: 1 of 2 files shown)


AnderGray:
This is very nice. Are you able to run the test in parallel on travis?

FriesischScott:
Yeah, but I like the integration with Github. It's easier to have everything in one platform. Also Travis can only do Linux and Mac. 

I did deactivate the tests for different platforms again, since we don't have any platform dependent code at the moment.

PR #18, opened 2020-08-19
Add Sobol indices and polyharmonic splines

FriesischScott:
This adds the following new features:

- Polyharmonic splines as a metamodel
- Computation of first order and total effect Sobol indices using Monte Carlo simulation
  - Quasi Monte Carlo methods return incorrect results for some reason
- Sobol sampling
- Halton sampling


FriesischScott:
@AnderGray Apart from the issue with QMC (that I can't figure out) I think this is ready to merge for now. Do you want to have a look or should I just merge it as is?

AnderGray:
It looks very nice!

We can take a look at the QMC at a future date.

PR #32, opened 2021-01-12
Add basic documentation

AnderGray:
Adds a template for GitHub documentation.

If all is well, then a docs page will be hosted in a gh-pages branch when the PR is merged.

Locally the API-docstrings says the functions are missing. I believe this because I'm using the registered version instead of the master branch. Not sure if this will be a problem in the online version.


FriesischScott:
The documentation has to be build against the branch and not the registered version.

FriesischScott:
@AnderGray I think we should just merge like this and then gradually improve on the documentation. Ideally we would do this now but I am way to lazy for that :smile:. 

I made a few changes:

- Removed related packages that we do not use yet. I don't see the point in name-dropping arbitrary packages.
- I also removed the bibliography. I think it would be better to cite the relevant literature at the appropriate place directly in the documentation, e.g. reference the line sampling paper when we explain how to run the simulation.
- A lot of restructuring and moving files into subfolders.

What do you think?

FriesischScott:
I don't exactly understand why `pull_request_target` fails `pull_request` fails, but I will remove the required check on the former.

AnderGray:
All the changes look good to me.

Happy to merge once we fix the `pull_request_target` fail

FriesischScott:
I'll remove `pull_request_target` in favor of `pull_request`. I tried the former to enable coverage stats for pull requests made from other repos. Didn't work though.

PR #34, opened 2021-02-09
Implement subset simulation

FriesischScott:
Adds subset simulation together with a few quality of life improvements here and there.


FriesischScott:
@AnderGray Mind taking a look before I finish up and add tests?

src/mcmc/metropolishastings.jl
```diff
@@ -0,0 +1,14 @@
+function metropolishastings(U0::AbstractMatrix, proposal::Sampleable{Univariate})
```
AnderGray, line 1: This looks quite strange to me. Is this a modified version of MH for subset simulation?

If so don't call it metropolishastings.

If it is the general algorithm, then there could be a couple of problems. 

- The acceptance probability is only right if the proposal is symmetric about the mean (the mv gaussian for example). 
- Why is your proposal only a 1D function
- The proposal sampler and the pdf are different

AnderGray (changes requested):
Generally looks great!

Small comment about the metropolis Hastings function.

src/mcmc/metropolishastings.jl
FriesischScott: You are right. I should probably just integrate it into the subset code since it's not valid metropolis-hastings anymore like this.

I am currently assuming the same proposal distribution for all variables. It should be easy to integrate different distributions, but I don't know if that is done in subset.

Everything is done in standard normal space so I don't need to pass a target distribution. It greatly simplifies the code. I'll check if that is actually correct. The results look decent though.

Do you know how I can assert that the proposal is symmetric? Just restrict it to certain distributions?

FriesischScott: I'll make some adjustments tomorrow.

AnderGray: There probably isn't an assertion that we can do, and it's probably fine for this.

The general acceptance prob is easy otherwise: 
`target(newSamp)/target(oldSamp) * proposal(oldSamp | newSamp) / proposal(newSamp | oldSamp)`

When it's symmetric: 
` proposal(oldSamp | newSamp) / proposal(newSamp | oldSamp) = 1`

AnderGray (approved):
Looks good! Happy to merge.

I like the example.

Does the subset work in multiple dimensions/ multiple performance functions? Or is it the case that you can merge several performances with max?

PR #54, opened 2021-12-05
Update compatability to 1.6 and 1.7

### LINKED ISSUE #53: Update compatability to 1.6 and 1.7
Now that 1.7 is released and 1.6 has become the new lts we should update the julia compatability to 1.6 and 1.7. Tests should also run on both versions. From now on I think it makes sense to always test on the lts and the latest stable.

FriesischScott:
Closes #53


FriesischScott:
The test solutions need to be updated because of the change in default rng.

FriesischScott:
Alternatively we could keep using mersenne-twister for the tests for the time being.

PR #57, opened 2022-01-11
Drop HaltonSequence dependency

Gopal1259:
The dependency on the HaltonSequence package is dropped. The Primes package has been implemented in order to calculate the prime numbers needed for sampling using Haltons Method.


FriesischScott:
You are still missing the updated `Project.toml`.

FriesischScott:
I contributed some changes to the QMC package to increase the accuracy with Float64. Please set the compat entry to `0.2.4` in the Project.toml.

test/Project.toml
```diff
@@ -1,6 +1,38 @@
+name = "UncertaintyQuantification"
```
FriesischScott, line 1: This is the wrong `Project.toml`. You need to update the one in the root folder...

FriesischScott (changes requested):

test/Project.toml
FriesischScott: Restore this to the way it was.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "UncertaintyQuantification"
 uuid = "7183a548-a887-11e9-15ce-a56ab60bad7a"
 authors = ["Jasper Behrensdorf <behrensdorf@irz.uni-hannover.de>", "Ander Gray <ander.gray@liverpool.ac.uk>"]
-version = "0.2.1"
+version = "0.2.4"
```
FriesischScott, line 4: This is the version of the UncertaintyQuantification package. You need to update the compat entry of QuasiMonteCarlo wich is lower in the file.

FriesischScott (changes requested):

FriesischScott (approved):

PR #63, opened 2022-04-08
Bayesian updating framework

FriesischScott:
This is where I'll update/improve the framework to better integrate it into the existing code base.


FriesischScott:
Implemented right now are

- Metropolis Hastings `mh()`
- Gibbs Sampler `gibbssample()`
- Transitional MCMC `tmcmc()`
- Sequential Monte Carlo `smc()`

along with some helper methods.

To clean this up I would like to unify the interface somehow. My suggestion:

Have all methods called `bayesian_inference()` or some variation of it. These will take the values that are the same among all algorithms while the extra parameters are grouped into structs like

```julia
struct TransitionalMarkovChainMonteCarlo begin
	...
end

struct SequentialMonteCarlo begin
	...
end
```

For metropolis-hastings as well as tmcmc we would just wrap around `TransitionalMCMC.jl` if possible. No need to reinvent the wheel.

FriesischScott:
Ideally, the method signature would include the arguments that all of the methods have (if they exist). Like
```julia
function bayesian_model_updating(likelihood::Function, prior::Function, bmu::AbstractBayesianModelUpdating)
    ...
end
```
where the `bmu` includes everything else based on the method that should be used (Transitional, Sequential, MH, etc.). If that makes sense.

Having a common interface and grouping the extras in the `struct` seems fairly reasonable. Not 100% sure though.

Project.toml
```diff
@@ -17,6 +17,7 @@ FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
FriesischScott, line 20: Please don't add a dependency on Plots.

src/inference/bayesianupdating.jl
```diff
@@ -0,0 +1,120 @@
...
+end
+
+function bayesianupdating(
+    loglikelihood::Function,
+    prior::Function,
+    amh::AdaptiveMetropolisHastings
+)
+    tar(x) = exp.(loglikelihood(x)).*prior(x)
+
+    x0 = [copy(amh.x0) for _ in 1:amh.cnum]
```
FriesischScott, line 50: Why create this if x0 is always the same?

src/inference/bayesianupdating.jl
```diff
@@ -0,0 +1,120 @@
...
+    gibbs::GibbsSampler
+)
+
+    error("Not implemented!")
+
+end
+
+# Adaptive Metropolis-Hastings, algorithm implemented according to Haario
+# et al., "An Adaptive Metropolis Algorithm", https://doi.org/10.2307/3318737
+function AdaptiveMetropolisHastingsSingle!(
```
FriesischScott, line 85: Functions should be lower- or snake-case.

src/inference/bayesianupdating.jl
jgrashorn: The AMH-implementation currently appends to the given DataFrame, so if multiple chains are desired, multiple DataFrames need to be created.

AnderGray:
For testing of MCMC methods in bayesian updating, we can use conjugate priors to get analytical solutions to simple updating problems.

There are tables in wikipedia: https://en.wikipedia.org/wiki/Conjugate_prior

For example a prior and likelihood which are both normal, will also produce a normal posterior. The parameters of the posterior depend on the parameters of the prior and likelihood. The updating rules can be found in the tables.

FriesischScott:
I will write a bit of documentation before the merge. Code -wise, it should be OK for now. We can always extend with more algorithms in the future.

AnderGray (approved):

src/modelupdating/bayesianupdating.jl
```diff
@@ -0,0 +1,259 @@
...
+end
+
+function bayesianupdating(
+    prior::Function, likelihood::Function, mh::SingleComponentMetropolisHastings
+)
+    return bayesianupdating(prior, likelihood, UQModel[], mh)
+end
+
+struct TransitionalMarkovChainMonteCarlo <: AbstractBayesianMethod # Transitional Markov Chain Monte Carlo
+    prior::Vector{RandomVariable}
```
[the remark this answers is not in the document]
AnderGray: Let's merge it for now, leave the branch open and we can come back later for further improvements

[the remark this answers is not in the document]
FriesischScott: This is because we need to sample from the prior in TMCMC where MH starts at a defined point.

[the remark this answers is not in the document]
FriesischScott: We can probably come up with  a better solution in the future.

PR #66, opened 2022-07-13
Split modified metropolis hastings  samples and evaluate cov in subset simulation

andreaperin:
The split is made in order to avoid ```evaluate!``` calls when ```candidatesamples``` generate one (or more) alpha-rejected sample.

### DIFF
--- a/src/simulations/subset.jl
+++ b/src/simulations/subset.jl
@@ -28,16 +28,11 @@ function sample(inputs::Array{<:UQInput}, sim::SubSetSimulation)
     if !isempty(deterministic_inputs)
         samples = hcat(samples, sample(deterministic_inputs, sim.n))
     end
-
     return samples
 end
 
-function probability_of_failure(
-    models::Union{Array{<:UQModel},UQModel},
-    performancefunction::Function,
-    inputs::Union{Array{T},T} where {T<:UQInput},
-    sim::SubSetSimulation,
-)
+function probability_of_failure(models::Union{Array{<:UQModel},UQModel},performancefunction::Function,inputs::Union{Array{T},T} where {T<:UQInput},sim::SubSetSimulation)
+
     random_inputs = filter(i -> isa(i, RandomUQInput), inputs)
     rvs = names(random_inputs)
 
@@ -48,11 +43,11 @@ function probability_of_failure(
     performance = [performancefunction(samples[end])]
 
     number_of_chains = Int64(max(1, ceil(sim.n * sim.target)))
-    samples_per_chain = floor(sim.n / number_of_chains)
+    samples_per_chain = Int64(floor(sim.n / number_of_chains))
 
     threshold = zeros(sim.levels, 1)
     pf = ones(sim.levels, 1)
-    CoV = zeros(sim.levels, 1)
+    cov = zeros(sim.levels, 1)
 
     for i in 1:(sim.levels)
         sorted_performance = sort(performance[end])
@@ -67,7 +62,7 @@ function probability_of_failure(
 
         ## Std MC covariance
         if i == 1
-            CoV[i] = sqrt((pf[i] - pf[i]^2) / (sim.n * pf[i]))
+            cov[i] = sqrt((pf[i] - pf[i]^2) / sim.n) / pf[i]
         end
 
         nextlevelsamples = [samples[end][sorted_indices[1:number_of_chains], :]]
@@ -79,9 +74,7 @@ function probability_of_failure(
 
             to_standard_normal_space!(inputs, chainsamples)
 
-            chainsamples[:, rvs], α_accept = candidatesamples(
-                Matrix{Float64}(chainsamples[:, rvs]), sim.proposal
-            )
+            chainsamples[:, rvs], α_accept = candidatesamples(Matrix{Float64}(chainsamples[:, rvs]), sim.proposal)
 
             to_physical_space!(inputs, chainsamples)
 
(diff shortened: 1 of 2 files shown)


src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
     accept = α .>= rand(size(α)...)
     θ[accept, :] = ξ[accept, :]
 
-    return θ
+    return θ, accept
+end
+
+function chaincovi(nextlevelperformance::Vector{Float64}, number_of_chains::Int64, samples_per_chain::Int64, threshold::Float64, pf::Float64, sim_n::Int64)
```
FriesischScott, line 148: Rename to `estimate_chain_cov`.

Passing `I` directly as a matrix removes the need for passing `number_of_chains` and `samples_per_chain` as they can be estimated from the matrix dimension`.

Rename `sim_n` to `n`.

src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
...
+    # Building indicator Matrix (Boolean)
+    indicator = reshape(nextlevelperformance, number_of_chains, samples_per_chain)
+    indicator = transpose(indicator .< max(threshold, 0))
+    ```Estimation of small failure probabilities in high dimensions by subset simulation
```
FriesischScott, line 152: Move reference to the paper into the docstring.

src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
...
+    ```
+    # Eq 29 - covariance vecotr between indicator(l) and indicator(l+k) -> ri
+    ri = zeros(1, samples_per_chain)
```
FriesischScott, line 155: Use subscript `i` for readability. 

Using `zeros(samples_per_chain` will give a `Vector` instead of a `Matrix`.

src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
...
+    for k in 1:samples_per_chain
+        for j in 1:number_of_chains
+            for l in 1:(Int(samples_per_chain - (k - 1)))
```
FriesischScott, line 158: Why do you need to cast to `Int` here?

src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
...
+                ri[k] = ri[k] + indicator[l, j] * indicator[l + k - 1, j]
+            end
+        end
+        ri[k] = ri[k] / (sim_n - (k - 1) * number_of_chains) - pf^2
+    end
+    # Eq 25 - correlation coefficient vector ρ
+    ρ = ri / ri[1]
+    # Eq 27 - γ_i Bernoulli coefficient 
+    γ_i = 0
```
FriesischScott, line 167: Something like this should work:

```julia
γᵢ = 2 * sum((1 .- (1:(samples_per_chain-1) .* number_of_chains ./ n)) .* ρ)
```

src/simulations/subset.jl
```diff
@@ -130,5 +142,35 @@ function candidatesamples(θ::AbstractMatrix, proposal::Sampleable{Univariate})
...
+    for k in 1:(samples_per_chain - 1)
+        γ_i = γ_i + (1 - k * number_of_chains / sim_n) * ρ[k]
+    end
+    γ_i = 2 * γ_i
+    #Eq 28 - i-level coefficient of variation (Metropolis Markov Chain)
+    cov_i = sqrt((1 - pf) / (pf * sim_n) * (1 + γ_i))
```
FriesischScott, line 173: Use δ like in the paper.

src/simulations/subset.jl
```diff
@@ -77,20 +74,26 @@ function probability_of_failure(
...
-            evaluate!(models, chainsamples)
-
-            chainperformance = performancefunction(chainsamples)
-
-            reject = chainperformance .> threshold[i]
-
-            chainperformance[reject] = nextlevelperformance[end][reject]
-            chainsamples[reject, rvs] = nextlevelsamples[end][reject, rvs]
+            ## Evaluating model just for new samples
+            α_accept_indices = findall(x -> x == true, α_accept)
```
FriesischScott, line 82: `findall(α_accept)`

src/simulations/subset.jl
```diff
@@ -77,20 +74,26 @@ function probability_of_failure(
...
+            if length(α_accept_indices) != 0
```
FriesischScott, line 83: `if any(α_accept)`

src/simulations/subset.jl
```diff
@@ -77,20 +74,26 @@ function probability_of_failure(
...
+                to_eval = chainsamples[α_accept, :]
+                evaluate!(models, to_eval)
+
+                to_evalperformance = performancefunction(to_eval)
+                performance_accept = to_evalperformance .< threshold[i]
+                chainsamples = copy(nextlevelsamples[end])
```
FriesischScott, line 89: Move lines 89 and 90 in front of if to remove duplication and drop the `else` block.

src/simulations/subset.jl
```diff
@@ -77,20 +74,26 @@ function probability_of_failure(
...
-            chainperformance = performancefunction(chainsamples)
-
-            reject = chainperformance .> threshold[i]
-
-            chainperformance[reject] = nextlevelperformance[end][reject]
-            chainsamples[reject, rvs] = nextlevelsamples[end][reject, rvs]
+            ## Evaluating model just for new samples
+            α_accept_indices = findall(x -> x == true, α_accept)
+            if length(α_accept_indices) != 0
+                to_eval = chainsamples[α_accept, :]
```
FriesischScott, line 84: Maybe find a better variable name than `to_eval`.

FriesischScott (changes requested):

FriesischScott (changes requested):

PR #69, opened 2022-07-27
Automatic deletion of external model files

andreaperin:
In order to not fill the hard disk during long simulation with external solvers, a _cleanup_ flag has been added to the struct _ExternalModel_. Its default value is set to _false_.
2 tests have been added in order to verify the functioning of the _cleanup_ flag.

### DIFF
--- a/test/models/externalmodel.jl
+++ b/test/models/externalmodel.jl
@@ -38,16 +38,15 @@
             sourcefiles,
             extrafiles,
             numberformats,
-            tempname(), 
+            tempname(),
             [r],
             opensees,
             false,
         )
         evaluate!(ext, df)
         @test length(readdir(readdir(ext.workdir; join=true)[1])) != 0
+        @test isapprox(df.r, sqrt.(df.x .^ 2 + df.y .^ 2))
     end
-    @test isapprox(df.r, sqrt.(df.x .^ 2 + df.y .^ 2))
-
 
     @testset "Cleanup" begin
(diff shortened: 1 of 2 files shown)


FriesischScott (changes requested):

test/models/externalmodel.jl
```diff
@@ -38,7 +32,37 @@
...
+            numberformats,
+            tempname(), 
+            [r],
+            opensees,
+            false,
+        )
+        evaluate!(ext, df)
+        @test length(readdir(readdir(ext.workdir; join=true)[1])) != 0
+    end
+    @test isapprox(df.r, sqrt.(df.x .^ 2 + df.y .^ 2))
```
FriesischScott, line 49: Should be inside the testset

test/models/externalmodel.jl
```diff
@@ -38,7 +32,37 @@
...
+            tempname(),
+            [r],
+            opensees,
+            true,
+        )
+        evaluate!(ext, df)
+        @test length(readdir(readdir(ext.workdir; join=true)[1])) == 0
+    end
 
     @test isapprox(df.r, sqrt.(df.x .^ 2 + df.y .^ 2))
```
FriesischScott, line 67: Should also be inside the testset

FriesischScott (changes requested):

FriesischScott (approved):

PR #93, opened 2023-01-06
Add Design of Experiments

mlsuh:
Adds a number of design of experiments to use with the `ResponseSurface`.

- TwoLevelFactorial
- FullFactorial
- FractionalFactorial
- CentralComposite
- BoxBehnken

The documentation and demo files include an example where the Himmelblau function is approximated.


FriesischScott:
Please move the `doe.jl` into the `simulations` subfolder.

mlsuh:
FractionalFactorial and BoxBehnken are now implemented. Can be tested by (un)commenting and manipulating the respective lines of code in ...demo/metamodels/responsesurface.jl

FriesischScott:
It turns out the mapping through standard normal space etc. is not the way to go. With the design of experiments we just want to cover the input space and don't care about the exact distributions. So we are going to do the following:

For finite bounds we just scale and shift the values between the bounds of the distribution. For example, for a `Uniform(-1, 1)` the values `[0.0, 0.5, 1.0]` would become`[-1, 0.0, 1.0]`. For infinite bounds we add another property to all the designs `σ::Integer=1`. In the transformation we will use `+` or `-` `σ` times the standard deviation of the rv as the bounds.

This way, no mappings through the `cdf` are performed and there is no risk of `inf` or `nan` values.

@mlsuh Can you adapt the `sample` function to behave like this?

FriesischScott (changes requested):
The tests don't pass.

src/simulations/doe.jl
```diff
@@ -0,0 +1,288 @@
...
+
+    # setting axial points
+    for i in 1:rvs
+        axial_points[(2i - 1):(2i), i] = [1.0, -1.0]
+    end
+
+    return vcat(two_lvl_points, axial_points)
+end
+
+function doe_samples(_::PlackettBurman, rvs::Int)
```
FriesischScott, line 281: Please don't commit work in progress functions.

src/simulations/doe.jl
```diff
@@ -0,0 +1,288 @@
...
+    end
+
+    return [ub lb]
+end
+
+function bounds(jd::JointDistribution, σ::Int)
+    return reduce(vcat, bounds.(jd.marginals, σ))
+end
+
+function sample(inputs::Array{<:UQInput}, design::AbstractDesignOfExperiments, σ::Int=1)
```
FriesischScott, line 60: The sigma should be a parameter of the individual structs of the designs, not the sample function.

FriesischScott (changes requested):

src/simulations/doe.jl
```diff
@@ -0,0 +1,288 @@
...
+
+struct PlackettBurman <: AbstractDesignOfExperiments end
+
+function full_factorial_matrix(levels::Vector{<:Integer})
+    ranges = [range(0.0, 1.0; length=l) for l in levels]
+    return mapreduce(t -> [t...]', vcat, Iterators.product(ranges...))
+end
+
+function bounds(r::RandomVariable, σ::Int)
+    ub = support(r.dist).ub
```
FriesischScott, line 42: `ub = maximum(rv)`

FriesischScott: `ub = isinf(ub) ? ub : std(rv) * sigma`

src/simulations/doe.jl
```diff
@@ -0,0 +1,288 @@
...
+
+    if ub == Inf
+        ub = std(r.dist) * σ
+    end
+
+    if lb == -Inf
+        lb = -std(r.dist) * σ
+    end
+
+    return [ub lb]
```
FriesischScott, line 53: return `[lb ub]`

src/simulations/doe.jl
mlsuh: what about `ub = isinf(maximum(rv)) ? std(rv) * sigma : maximum(rv)`?
So, one line and no refactoring of ub but with tradeoff of extra call of `maximum(rv)`.

FriesischScott (approved):

FriesischScott:
Nice work @mlsuh!

PR #98, opened 2023-03-02
Add multiple outputs for the evaluation of Sobol indices

andreaperin:
- the 'output' argument in the 'sobolindices' funcition have to be a vector

### DIFF
--- a/src/sensitivity/sobolindices.jl
+++ b/src/sensitivity/sobolindices.jl
@@ -78,3 +78,78 @@ function sobolindices(pce::PolynomialChaosExpansion)
 
     return indices
 end
+
+function sobolindices(
+    models::M where {M<:UQModel},
+    inputs::Vector{<:UQInput},
+    outputs::Vector{Symbol},
+    sim::AbstractMonteCarlo,
+)
+    models = [models]
+    return sobolindices(models, inputs, outputs, sim)
+end
+
+function sobolindices(
+    models::Vector{<:UQModel},
+    inputs::I where {I<:UQInput},
+    outputs::Vector{Symbol},
+    sim::AbstractMonteCarlo,
+)
+    inputs = [inputs]
+    return sobolindices(models, inputs, outputs, sim)
+end
+
+function sobolindices(
+    models::Vector{<:UQModel},
+    inputs::Vector{<:UQInput},
+    outputs::Symbol,
+    sim::AbstractMonteCarlo,
+)
+    outputs = [outputs]
+    return sobolindices(models, inputs, outputs, sim)
+end
+
+function sobolindices(
+    models::M where {M<:UQModel},
+    inputs::I where {I<:UQInput},
+    outputs::Vector{Symbol},
+    sim::AbstractMonteCarlo,
+)
+    models = [models]
+    inputs = [inputs]
+    return sobolindices(models, inputs, outputs, sim)
+end
+
+function sobolindices(
+    models::M where {M<:UQModel},
+    inputs::Vector{<:UQInput},
+    outputs::Symbol,
+    sim::AbstractMonteCarlo,
+)
+    models = [models]
+    outputs = [outputs]
+    return sobolindices(models, inputs, outputs, sim)
+end
+
+function sobolindices(
+    models::Vector{<:UQModel},
+    inputs::I where {I<:UQInput},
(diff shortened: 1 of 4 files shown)


src/sensitivity/sobolindices.jl
```diff
@@ -13,50 +13,53 @@ function sobolindices(
 
     evaluate!(models, samples)
 
-    A = samples[1:(sim.n), :]
-    B = samples[(sim.n + 1):end, :]
+    indices = Dict()
+    for quantity in output
```
FriesischScott, line 17: By putting the loop here you are reevaluating the model for every output. Move the loop inside the indices computation to avoid this.

FriesischScott: Also quantity is not a good variable name.

Just go with outputs and output or something like that.

src/sensitivity/sobolindices.jl
```diff
@@ -13,50 +13,53 @@ function sobolindices(
...
-    indices.TotalEffectStdError = STi[:, 2]
+        indices[quantity] = DataFrame()
+        indices[quantity].Variables = random_names
+        indices[quantity].FirstOrder = Si[:, 1]
+        indices[quantity].FirstOrderStdError = Si[:, 2]
+        indices[quantity].TotalEffect = STi[:, 1]
+        indices[quantity].TotalEffectStdError = STi[:, 2]
+    end
 
     return indices
```
FriesischScott, line 64: For a single output it should return the `DataFrame` directly and not wrapped in a `Dict`.

test/sensitivity/sobolindices.jl
```diff
@@ -13,45 +13,45 @@
...
-        si = sobolindices([ishigami], x, :f, MonteCarlo(n))
+        si = sobolindices([ishigami], x, [:f], MonteCarlo(n))
 
         Random.seed!()
 
-        @test all(isapprox(si.FirstOrder, firstorder_analytical; rtol=0.1))
-        @test all(isapprox(si.TotalEffect, totaleffect_analytical; rtol=0.1))
+        @test all(isapprox(si[:f].FirstOrder, firstorder_analytical; rtol=0.1))
+        @test all(isapprox(si[:f].TotalEffect, totaleffect_analytical; rtol=0.1))
     end
```
FriesischScott, line 55: Can you put a simple example with two models and outputs to make sure it works?

src/sensitivity/sobolindices.jl
FriesischScott: You could use different values for `a` and `b` and create a second ishigami model. Then look up/ calculate the analytical values for the second one as well.

FriesischScott:
It's probably a good idea to save the output values in a Matrix and then do the computations for all from there.

FriesischScott (changes requested):
Almost done.

src/sensitivity/sobolindices.jl
```diff
@@ -12,20 +17,23 @@ function sobolindices(
     random_names = names(filter(i -> isa(i, RandomUQInput), inputs))
 
     evaluate!(models, samples)
+    indices = Dict([
+        (name, DataFrame(_sobol_table_types, _sobol_table_header)) for name in outputs
+    ])
+
+    if length(outputs) == 1
```
FriesischScott, line 24: Don't do this.

src/sensitivity/sobolindices.jl
```diff
@@ -35,29 +43,28 @@ function sobolindices(
...
+            Sᵢ = first_order(ABi[:, qty])
+            σSᵢ = stderror(bs)[1]
+
+            # Total effects
+            total_effect = x -> (1 / (2 * sim.n)) * sum((fA[:, j] .- x) .^ 2) / VY[j] # Saltelli 2009
+            bs = bootstrap(total_effect, ABi[:, qty], BasicSampling(1000))
+            Sₜ = total_effect(ABi[:, qty])
+            σSₜ = stderror(bs)[1]
+
+            if length(outputs) == 1
```
FriesischScott, line 61: Do this

```julia
return length(outputs) > 1 ? indices : indices[outputs] # might need outputs[1] not sure
```

FriesischScott (changes requested):

PR #103, opened 2023-03-31
Replace Symbolics with recurrence formulas

FriesischScott:
Turns out this is much faster and cleaner than using `Symbolics` to build functions. As a benefit the dependency (which is quite heavy) can be dropped altogether.

The constructor for the pce bases changed slighty as a result. The degree doesn't need to be included any more.


FriesischScott:
The execution time goes from milliseconds to microseconds. Pretty impressive.

PR #121, opened 2023-07-19
Fix solver for relative workdir

andreaperin:
Add a check on solver path.
        
- stdout and stderr from solver run do not depend anymore on relative paths
- a warning will be throw when he solver has a relative path

### DIFF
--- a/src/solvers/solver.jl
+++ b/src/solvers/solver.jl
@@ -2,6 +2,12 @@ struct Solver
     path::String
     source::String
     args::String
+    function Solver(path::String, source::String, args::String)
+        if !isabspath(path)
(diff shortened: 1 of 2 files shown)


src/solvers/solver.jl
```diff
@@ -2,6 +2,12 @@ struct Solver
     path::String
     source::String
     args::String
+    function Solver(path::String, source::String, args::String)
+        if !isabspath(path)
+            @warn "Specified solver path is not an absolute path. Make sure $path is in your environmental varibles path."
```
FriesischScott, line 7: Solver path is not an absolute path. Make sure $path is on your PATH.

FriesischScott (changes requested):

FriesischScott (approved):

PR #124, opened 2023-08-02
Add importance sampling

Cr0gan:
Importance sampling works so far for the cantilever example (in: ./demo/IS_test.jl).
I have attached functions that evaluate the pdf for samples in a DataFrame.

Commented out are currently:
 - the sampling method from Matteo's lecture (in: src/simulations/importancesampling.jl)
 - the corresponding probability_of_failure method (in: src/reliability/probabilityoffailure.jl)
 - a function to create user-defined proposal distributions (in: src/simulations/importancesampling.jl)


FriesischScott:
English please :).

src/reliability/form.jl
```diff
@@ -65,5 +66,5 @@ function probability_of_failure(
     to_physical_space!(inputs, dp)
 
     dp = (; zip(random_names, collect(dp[1, :]))...)
-    return pf, β, dp
+    return pf, β, dp, α
```
FriesischScott, line 69: Needs an update to the test to verify the important direction.

src/reliability/probabilityoffailure.jl
```diff
@@ -65,6 +65,47 @@ function probability_of_failure(
...
+#     cov = sqrt(variance) / pf
+
+#     return pf, cov, samples
+# end
+
+function probability_of_failure(
+    models::Union{Vector{<:UQModel},UQModel},
+    performance::Function,
+    inputs::Union{Vector{<:UQInput},UQInput},
+    proposals::Union{Vector{<:UQInput},UQInput},
```
FriesischScott, line 92: The general signature for `probability_of_failure` must not change.

src/simulations/importancesampling.jl
```diff
@@ -0,0 +1,102 @@
+struct ImportanceSampling
+    sampling::AbstractMonteCarlo
+end
+
+# function sample(inputs::Vector{<:UQInput}, models::Union{Vector{<:UQModel},UQModel}, performance::Function, design_point_from::FORM, sim::ImportanceSampling)
```
FriesischScott, line 5: I discussed with Matteo and this is the only way to do importance sampling we need to include for now. The normals centered around the design point are inefficient.

FriesischScott (changes requested):
Sorry for this 😅 . But please remove all the proposal stuff and only implement the method based on the `Z` matrix using the design point and important direction which is currently commented out.

Put all necessary information into the `ImportanceSampling` struct. The parameter `c` can have a default of 2.

Maybe we should have an alternative implementation where `FORM` is used automatically to get the design point on the fly.

src/inputs/parameter.jl
```diff
@@ -22,3 +22,4 @@ to_standard_normal_space!(p::Parameter, x::DataFrame) = nothing
 to_physical_space!(p::Parameter, x::DataFrame) = nothing
 
 mean(p::Parameter) = p.value
+pdf(p::Parameter, x::DataFrame) = x[!, p.name] ./ x[!, p.name]
```
FriesischScott, line 25: First of all this is 1.0. And second I don't think you use it anywhere.

src/inputs/randomvariable.jl
```diff
@@ -91,4 +91,4 @@ minimum(rv::RandomVariable) = minimum(rv.dist)
 maximum(rv::RandomVariable) = maximum(rv.dist)
 insupport(rv::RandomVariable, x::Real) = insupport(rv.dist, x)
 mean(rv::RandomVariable) = mean(rv.dist)
-var(rv::RandomVariable) = var(rv.dist)
+var(rv::RandomVariable) = var(rv.dist)
```
FriesischScott, line 94: Revert this so it doesn't show up as a change.

src/UncertaintyQuantification.jl
```diff
@@ -94,6 +95,7 @@ export mean
 export multivariate_indices
 export polynomialchaos
 export probability_of_failure
+export proposaldistribution
```
FriesischScott, line 98: Leftover, can be removed.

demo/reliability/cantilever.jl
```diff
@@ -34,7 +34,7 @@ displacement = Model(
 max_displacement = 0.01
 
 # Compute probability of failure using standard Monte Carlo
-mc = MonteCarlo(10^6)
+mc = MonteCarlo(10^3)
```
FriesischScott, line 37: Please revert this back for an accurate solution.

demo/reliability/cantilever.jl
```diff
@@ -44,6 +44,18 @@ println(
     "Monte Carlo probability of failure $mc_pf ($(size(mc_samples, 1)) model evaluations)"
 )
 
+# Compute probability of failure using Importance Sampling
+pf, β, dp, α = probability_of_failure([inertia, displacement], df -> max_displacement .- df.w, inputs, FORM())
+is = ImportanceSampling(1000, β, dp, α)
```
FriesischScott, line 49: Maybe use `1e4` for a comparable solution to standard MC.

test/reliability/form.jl
```diff
@@ -8,12 +8,12 @@
 
     model = Model(df -> df.x1 .- df.x2, :y)
 
-    pf, β, dp = probability_of_failure(df -> df.x1 .- df.x2, inputs, form)
+    pf, β, dp, α = probability_of_failure(df -> df.x1 .- df.x2, inputs, form)
```
FriesischScott, line 11: Verify alpha is correct.

test/reliability/form.jl
```diff
@@ -8,12 +8,12 @@
...
 
     @test round(pf; digits=4) ≈ 0.0127
     @test round(β; digits=4) ≈ 2.2361
 
-    pf, β, dp = probability_of_failure(model, df -> df.y, inputs, form)
+    pf, β, dp, α = probability_of_failure(model, df -> df.y, inputs, form)
```
FriesischScott, line 16: Verify alpha is correct.

FriesischScott (changes requested):
Looking very good apart from some very minor issues.

FriesischScott:
After removing the `pdf` of the `Parameter` patch coverage should also be 100%.

src/reliability/probabilityoffailure.jl
```diff
@@ -65,6 +65,25 @@ function probability_of_failure(
...
+    sim::ImportanceSampling
+)
+    samples, weights = sample(inputs, sim)
+    evaluate!(models, samples)
+
+    # Probability of failure
+    weighted_failures = (performance(samples) .< 0) .* weights.w
+    pf = sum(weighted_failures) / sim.n
+
+    variance = ((sum(weighted_failures .* weights.f) / sim.n) - pf^2) / sim.n
```
FriesischScott, line 81: I think this should be `weights.h^2/weights.f`. This is from Matteos lecture:

```Matlab
var_pf = 1/Nsamples*(sum(failure.*normpdf(Zforced,0,1).^2./normpdf(Zforced,b,v))/Nsamples - pf^2); 
```

FriesischScott (changes requested):

src/reliability/probabilityoffailure.jl
FriesischScott: This will possibly fix the issue with slightly negative variance.

Cr0gan: In the sample method `weights.f` is `normpdf(Zforced, 0, 1)`, `weights.h` is `normpdf(Zforced, b, v)` and `weights.w` is `weights.f ./ weights.h`. Therefore `weights.w .* weights.f` should be right imo. 

Could the issue be due to some precision errors, or is the variance's absolute value too high for that?

teobros: It should be this
`variance = ((sum(weighted_failures .* prod(normpdf(samples)) / sim.n) - pf^2) / sim.n`
but samples are in SNS for this formula

Cr0gan: Alright, I can change that. But wouldn't that potentially lead to a negative variance as well? I think `prod(normpdf(samples))` would essentially go to zero for large amounts of samples and we would end up with `(0 - pf^2) / sim.n`.

FriesischScott: I don't think it will. You need to essentially compute the product of the columns of the pdf values.

FriesischScott (approved):

FriesischScott:
Great stuff @Cr0gan, thanks!

PR #128, opened 2023-08-29
Fix mse for least squares PCE

andreaperin:
The evaluation of the mse of PCE with "least squared" method was performed without the division by the number of samples.


FriesischScott:
Can you rebase you commit on `master` so the history is correct? The solver stuff was already merged in a previous PR.

test/models/pce/polynomialchaosexpansion.jl
```diff
@@ -16,10 +16,11 @@
 
     @testset "LeastSquares" begin
         ls = LeastSquares(SobolSampling(1000))
-        pce, _, _ = polynomialchaos(x, model, Ψ, :y, ls)
+        pce, _, mse = polynomialchaos(x, model, Ψ, :y, ls)
 
         @test mean(pce) ≈ -1.5 rtol = 1e-10
         @test var(pce) ≈ 0.5 rtol = 1e-10
+        @test isapprox(mse, 3e-30, atol=1e-10)
```
FriesischScott, line 23: This is not an adequate test for the mse. You need to get the `y` values from the returned `samples` and compare them to `evaluate!(pce, copy_of_samples)` with `copy_of_samples = samples[:, Not(:y)]`.

FriesischScott (changes requested):

PR #129, opened 2023-09-12
Add basic empirical distribution

FriesischScott:
The distribution is fit on data by using kernel density estimation for the PDF. The CDF and quantile function are computed empirically and using spline interpolation.

The distribution is implemented as a subtype of `ContinuousUnivariateDistribution` and should work smoothely in the `RandomVariable` and everywhere else.


FriesischScott:
@andreaperin Please confirm this actually works.

Issue #138, opened 2023-10-26
estimate_cov for subset and subset infinity sometimes returns complex number


AnderGray:
I'm a convergence study for MC, subset, and subset infinity.

Sometimes (randomly) `estimate_cov` is trying to take the sqrt of a complex number. Script for the study

```julia
using UncertaintyQuantification, DelimitedFiles, Formatting
using Distributions, Plots

###
#   Define input random variables
###
X1 = RandomVariable(Normal(0, 1), :X1)
X2 = RandomVariable(Normal(0, 1), :X2)

inputs = [X1, X2]

###
#   Define function, and model
###
function y(df)
    return df.X1 .+ df.X2
end

m = Model(y, :y)

###
#   Define limitstate. Failure is limitstate <= 0
###
function limitstate(df)
    return 4 .- reduce(vcat, df.y)
end

###
#   True failure for this problem
###
True_failure_prob = 1 - cdf(Normal(), 4/sqrt(2))

N_samples = 10 .^range(2, 5, 50)

pf_1 = zeros(length(N_samples))
pf_2 = zeros(length(N_samples))
pf_3 = zeros(length(N_samples))

cov_1 = zeros(length(N_samples))
cov_2 = zeros(length(N_samples))
cov_3 = zeros(length(N_samples))


for (i, N) in enumerate(N_samples)
    println(i)
    ###
    #   Try different simulation methods
    ###
    simulation_method_1 = MonteCarlo(Integer(round(N * 3)))
    simulation_method_2 = SubSetSimulation(Integer(round(N)), 0.1, 10, Normal())
    simulation_method_3 = SubSetInfinity(Integer(round(N)), 0.1, 10, 0.5)

    ###
    #   Simulate
    ###
    pf_1[i], cov_1[i], _ = probability_of_failure(m, limitstate, inputs, simulation_method_1)
    pf_2[i], cov_2[i], _ = probability_of_failure(m, limitstate, inputs, simulation_method_2)
    pf_3[i], cov_3[i], _ = probability_of_failure(m, limitstate, inputs, simulation_method_3)

end



plot(N_samples, pf_1)
plot!(N_samples, pf_2)
plot!(N_samples, pf_3)
plot!(N_samples, ones(length(N_samples)) * True_failure_prob,  xaxis=:log10)


plot(N_samples, cov_1)
plot!(N_samples, cov_2)
plot!(N_samples, cov_3, xaxis=:log10)
```

```
[...]

PR #140, opened 2023-11-15
Adds adaptive subset infinity, and subset simulation logging

AnderGray:
An implementation of adaptive conditional subset simulation (subset infinity) from [1,2].

The idea behind the algorithm is to adaptively tune the standard deviation (`s` in our implementation) of the proposal distribution towards an optimal acceptance rate. Which they say is `a_star = 0.44`. 

For each level, the algorithm iteratively computes the acceptance rate. This is done by partitioning the seeds (say into 4), and running usual subset infinity 4 times, and computing a scaling factor `λ`, which updates the std `s_new = min(1, λ*s)`. In this case, `s` is updated 4 times for each level.

Here's my data structure:

```Julia
mutable struct SubSetInfinityAdaptive <: AbstractSubSetSimulation
    n::Integer
    target::Float64
    levels::Integer
    λ::Real
    Na::Integer   
    s::Real
end
```

The new properties: 
*  `λ`: initial scaling factor. Recommended to be set to `1`
* `Na`: the number of times to update the scaling factor in each level. The number of seeds is partitioned this many times

If `Na=1`, the update will only happen at each level.

Note, in my implementation the structures has to be mutable, since I'm updating and saving `λ` at each level. Perhaps there's a better way, but this way I was able to use the same `probability_of_failure` function.

Here's the constructor:

```Julia
    function SubSetInfinityAdaptive(
        n::Integer, target::Float64, levels::Integer, λ::Real, Na::Integer
    ) 
        number_of_seeds = Int64(max(1, ceil(n * target)))

        (Na <= number_of_seeds) ||
            error("Number of partitions Na must be less than `n` * `target`")
        (mod(number_of_seeds, Na) == 0) ||
            error("Number of partitions Na must be a multiple of `n` * `target`")
        (0 <= λ <= 1) ||
            error("Scaling parameter must be between 0.0 and 1.0. A good initial choice is 1.0")
        return new(n, target, levels, 1, Na, 1)
    end
```

The second check is important to ensure the seeds can be partitioned into `Na`.

### Further future improvements

* Have a different scaling factor for each dimension. The papers state that the sample std and mean can be computed between levels, allowing for the proposal's variance to be different in each dimension. This allows for sensitive dimensions to be determined by algorithm.

### LOGGING

It would be good to **Logging.jl** the progress of the subset simulation algorithms. I've been running quite expensive simulations, and it's useful to print progress. I've added some, however I think they can be improved... Currently too much information is being printed.

### Refs

[1] Papaioannou, I., Betz, W., Zwirglmaier, K., & Straub, D. (2015). MCMC algorithms for subset simulation. Probabilistic Engineering Mechanics, 41, 89-103.

[2] Chan, J., Papaioannou, I., & Straub, D. (2022). An adaptive subset simulation algorithm for system reliability analysis with discontinuous limit states. Reliability Engineering & System Safety, 225, 108607.

### DIFF
--- a/src/simulations/subset.jl
+++ b/src/simulations/subset.jl
@@ -89,7 +89,6 @@ function sample(inputs::Vector{<:UQInput}, sim::AbstractSubSetSimulation)
     return samples
 end
 
-
 """
     SubSetInfinityAdaptive(n::Integer, target::Float64, levels::Integer, ....)
 
@@ -124,14 +123,16 @@ mutable struct SubSetInfinityAdaptive <: AbstractSubSetSimulation
     levels::Integer
     λ::Real
     Na::Integer     # Probably needs to be a multiple of N
-    s :: Real
+    s::Real
 
-    function SubSetInfinityAdaptive(n::Integer, target::Float64, levels::Integer, λ::Real, Na::Integer)
-        (0 <= λ <= 1) || error("Scaling parameter must be between 0.0 and 1.0. A good initial choice is")
+    function SubSetInfinityAdaptive(
+        n::Integer, target::Float64, levels::Integer, λ::Real, Na::Integer
+    )
+        (0 <= λ <= 1) ||
+            error("Scaling parameter must be between 0.0 and 1.0. A good initial choice is")
         return new(n, target, levels, 1, Na, 1)
     end
 end
-
 
 function probability_of_failure(
     models::Union{Vector{<:UQModel},UQModel},
@@ -327,8 +328,6 @@ function nextlevelsamples(
     return nextlevelsamples, nextlevelperformance
 end
 
-
-
 function nextlevelsamples(
     samples::DataFrame,
     performance::Vector{<:Real},
@@ -337,11 +336,9 @@ function nextlevelsamples(
     performancefunction::Function,
     inputs::Union{Vector{<:UQInput},UQInput},
     sim::SubSetInfinityAdaptive,
-)   
-
+)
     next_samples = [samples]
     next_performance = [performance]
-
 
     a_star = 0.44       # Optimal acceptance rate
     N_iter = 4          # Number of iterations or updates of s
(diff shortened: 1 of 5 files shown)


src/simulations/subset.jl
```diff
@@ -89,6 +89,58 @@ function sample(inputs::Vector{<:UQInput}, sim::AbstractSubSetSimulation)
...
+# Examples
+
+```jldoctest
+julia> SubSetInfinityAdaptive(100, 0.1, 10, 1, 2)
+SubSetInfinityAdaptive(100, 0.1, 10, 1, 2)
+```
+
+# References
+
+[papaioannou2015mcmc](@cite)
```
FriesischScott, line 117: These need the matching entries in the bibfile.

src/simulations/subset.jl
```diff
@@ -89,6 +89,58 @@ function sample(inputs::Vector{<:UQInput}, sim::AbstractSubSetSimulation)
...
+Idea behind this algorithm is to adaptively select the correlation parameter of `s`
+at each intermediate level, by simulating a subset N_a of the chains
+(which must be choosen without replacement at random) and modifying the acceptance rate towards the optiming
+α_star = 0.44
+
+# Examples
+
+```jldoctest
+julia> SubSetInfinityAdaptive(100, 0.1, 10, 1, 2)
+SubSetInfinityAdaptive(100, 0.1, 10, 1, 2)
```
FriesischScott, line 112: The output for the doctest needs to be `SubSetInfinityAdaptive(100, 0.1, 10, 1, 2, 1)` because of the default `s` added in the constructor.

src/simulations/subset.jl
AnderGray: Thanks, added the references.

Btw, the references don't print in the Julia help.

I.e.: `? SubSetSimulation` prints `auEstimationSmallFailure2001 (@cite)` and not the full citation. Is this normal behaviour?

FriesischScott: Yes, the citations are only part of `Documenter.jl`.

test/reliability/probabilityoffailure.jl
```diff
@@ -75,5 +75,13 @@
 
         # 95% conf intervals estimated from 1000 runs
         @test 1.8e-11 < pf < 1.6e-9
+    
+        subset = SubSetInfinityAdaptive(10^4, 0.1, 20, 4)
```
FriesischScott, line 79: Since we're already touching this I think it would be good to put the subset tests into separate test sets. The test is still somewhat unstable and this way it's easier to see which specific subset implementation causes a failure.

The setup for the tests can be done outside the testsets. Like this inside the larger probability of failure testset.

```julia
# Kontantin Zuev - Subset Simulation Method for Rare Event Estimation: An Introduction
# Example 6.1
# Target pf of 1e-10
pf_analytical = 1e-10

x1 = RandomVariable(Normal(), :x1)
x2 = RandomVariable(Normal(), :x2)

y = Parameter(sqrt(2) * quantile(Normal(), 1 - pf_analytical), :y)

g = Model(df -> df.x1 + df.x2, :g)

F = df -> df.y .- df.g

@testset "Subset Simulation" begin
    subset = SubSetSimulation(10^4, 0.1, 20, Normal())

    pf, _, _ = probability_of_failure(g, F, [x1, x2, y], subset)

    # 95% conf intervals estimated from 1000 runs
    @test 1.4e-11 < pf < 9.1e-10
end

@testset "Subset Infinity" begin
    subset = SubSetInfinity(10^4, 0.1, 20, 0.5)

    pf, _, _ = probability_of_failure(g, F, [x1, x2, y], subset)

    # 95% conf intervals estimated from 1000 runs
    @test 1.8e-11 < pf < 1.6e-9
end

@testset "Subset Infinity Adaptive" begin
    subset = SubSetInfinityAdaptive(10^4, 0.1, 20, 1, 4)

    pf, _, _ = probability_of_failure(g, F, [x1, x2, y], subset)

    # 95% conf intervals estimated from 1000 runs
    @test 3.14e-11 < pf < 3.4e-10
end
```

docs/references.bib
```diff
@@ -83,3 +83,22 @@ @incollection{zuevSubsetSimulationMethod2013
   langid     = {english}
 }
 
+@article{papaioannou2015mcmc,
```
FriesischScott, line 86: Sorry to be annoying, but please indent and alphabetically sort the bibfile. The LaTeX Workshop extension for vscode can do that for you.

test/reliability/probabilityoffailure.jl
AnderGray: Nice idea

FriesischScott:
Thanks for the PR, looking very nice overall!

The only thing I don't care for is the logging. It's very noisy as it is now.

I would prefer it if we remove it for now and make a larger PR for nice logging in more places then just subset.

AnderGray:
Thanks for the review!

Yes, I agree about the logging. However, I am finding it useful when running expensive models. For now, would it work if we replace the `@info` with `@debug`?

The other thing we could consider adding is an example of an ablation test for all the subset methods. Something that plots the result as a function of input `dimension` and input samples `N`. Something similar to what's done in this example: #138. I have a script ready which does this. Can also add later once we fix that issue.

FriesischScott:
I'd be fine with using `@debug` for now.

AnderGray:
Looks good from my side.

What do you think about having an option without using Markov-chain, like we had it before?

FriesischScott:
Thanks, I'm merging for now.

> What do you think about having an option without using Markov-chain, like we had it before?

Does that make sense though? Does it still sample from the correct conditional pdf? Maybe we should run some benchmarks.

PR #143, opened 2023-11-20
Add randomized QMC

### LINKED ISSUE #144: Randomized QMC: tests
Currently, there are only tests for unrandomized qmc. Tests for randomized Sobol-, Halton- and Lattice Rule Sampling will need to be implemented.

mlsuh:
Using methods from QuasiMonteCarlo.jl

Closes #144.


src/simulations/montecarlo.jl
```diff
@@ -5,12 +5,54 @@ end
...
+        base::Integer=2,
+        pad::Integer=32,
+    )
+        randomization ∉ [:matousekscramble, :digitalshift, :shift, :owenscramble, :none] &&
+            error(
+                "type must be :matousekscramble, :digitalshift, :shift, :owenscramble or :none",
+            )
+        pad < log(base, n) && error("pad must be ≥ log(base, n)")
+        if n > 0
+            if !isinteger(log(base, n))
```
FriesischScott, line 47: The number of samples being a power of 2 is only relevant for `SobolSampling`.

src/simulations/montecarlo.jl
```diff
@@ -20,7 +62,31 @@ end
...
```
FriesischScott, line 81: See above, not valid for `LatticeRuleSampling`.

FriesischScott (changes requested):

src/simulations/montecarlo.jl
mlsuh: Just noticed that the warning message (l. 49) is not correct, it should be `$base` instead of `2`.

FriesischScott:
Based on the feedback from the QMC.jl devs we need to restrict lattice rules to shifting and allows scrambling only for digital nets (sobol, faure (to be added)).

Then we should contribute the `rhalton` sampling to the QMC package to allow for scrambled halton sequences. https://arxiv.org/pdf/1706.02808.pdf

Project.toml
```diff
@@ -21,6 +18,7 @@ Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
 KernelDensity = "5ab0869b-81aa-558d-bb23-cbf5423bbe9b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
+Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
```
FriesischScott, line 21: Needs compat entry.

src/simulations/montecarlo.jl
```diff
@@ -132,13 +125,13 @@ function qmc_samples(sim::LatticeRuleSampling, rvs::Integer)
...
 
-function randomize(sim::AbstractQuasiMonteCarlo, u::Matrix)
+function randomize(sim::AbstractQuasiMonteCarlo, u::Matrix, b=2)
     if sim.randomization == :matousekscramble
-        u = QuasiMonteCarlo.randomize(u, MatousekScramble(; base=sim.base, pad=sim.pad))
+        u = QuasiMonteCarlo.randomize(u, MatousekScramble(; base=b))
     elseif sim.randomization == :owenscramble
-        u = QuasiMonteCarlo.randomize(u, OwenScramble(; base=sim.base, pad=sim.pad))
+        u = QuasiMonteCarlo.randomize(u, OwenScramble(; base=b))
     elseif sim.randomization == :digitalshift
```
FriesischScott, line 133: Drop :digitalshift.

FriesischScott (changes requested):
Please sync with `master` and add test to make sure the randomization returns actual random points.

FriesischScott (approved):

PR #146, opened 2023-12-11
Imprecise probabilities

andreaperin:
Implementation of imprecise probabilities with external double loop by mean of:

1. Interval struc
2. ProbabilityBox struc (no distribution-free)
3. new method for ```probability_of_failure``` function


Project.toml
```diff
@@ -21,6 +18,7 @@ Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
 KernelDensity = "5ab0869b-81aa-558d-bb23-cbf5423bbe9b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
+PRIMA = "0a7d04aa-8ac2-47b3-b7a7-9dbd6ad661ed"
```
FriesischScott, line 21: Needs a compat entry.

src/UncertaintyQuantification.jl
```diff
@@ -27,10 +28,12 @@ import Distributions: logpdf, pdf, cdf, quantile, minimum, maximum, insupport, m
 abstract type UQType end
 
 abstract type UQInput <: UQType end
+abstract type PreciseUQInput <: UQInput end
```
FriesischScott, line 31: Not fully convinced by this terminology. Maybe we can find something more fitting.

test/reliability/probabilityoffailure.jl
```diff
@@ -1,79 +1,113 @@
...
+
+        subset = SubSetInfinity(10^4, 0.1, 20, 0.5)
+
+        pf, _, _ = probability_of_failure(g, F, [x1, x2, y], subset)
+
+        # 95% conf intervals estimated from 1000 runs
+        @test 1.8e-11 < pf < 1.6e-9
+    end
+
+    @testset "Imprecise Probabilities Simulation" begin
```
FriesischScott, line 80: Let's not use this example here but a simpler one.

FriesischScott (changes requested):
Still missing the second algorithm ;).

Please fix the conflicts and sync with `master`.

Once the second algorithm is here I would like to see a comparison of the three global optimization toolboxes we discussed: PRIMA, BlackBoxOptim and NOMAD.

FriesischScott:
@AnderGray This is still very much in progress, but if you have time and want to provide some feedback please go ahead.

src/reliability/probabilityoffailure.jl
```diff
@@ -1,92 +1,145 @@
...
+)
+    return probability_of_failure(UQModel[], performance, wrap(inputs), sim)
+end
+
+# ImpreciseUQInput functions
+function probability_of_failure(
+    models::Union{Vector{<:UQModel},UQModel},
+    performance::Function,
+    inputs::Union{Vector{<:UQInput},UQInput},
+    sim::AbstractMonteCarlo,
```
AnderGray, line 99: Very cool. If you replace with `Union{AbstractMonteCarlo, AbstractSubSetSimulation}` it 'just works' with subset sim too.

AnderGray: I think the same applies to other simulation methods if changed to `Any`

src/UncertaintyQuantification.jl
AnderGray: I agree. Also it is probably better to have a new type for the imprecise inputs rather than the precise ones? 

Since most of the package uses precise probability, and we only need to dispatch on the new imprecise stuff.

src/inputs/imprecise/p-box.jl
```diff
@@ -0,0 +1,38 @@
...
+```jldoctest
+julia>  ProbabilityBox([1.75, 1.83], [1.77, 1.85], x -> Uniform(x...), :l)
+ProbabilityBox(Real[1.75, 1.83], Real[1.77, 1.85], var"#7#8"(), :l)```
+"""
+struct ProbabilityBox <: ImpreciseUQInput
+    lb::AbstractVector{<:Real}
+    ub::AbstractVector{<:Real}
+    dist::Function
+    name::Symbol
+    function ProbabilityBox(lb::AbstractVector{<:Real}, ub::AbstractVector{<:Real}, dist::Function, name::Symbol)
```
AnderGray, line 22: The constructor is slightly confusing ...

I think would be ideal to construct the p-box where you specify the distribution parameters as intervals, rather than making a vector of lower and upper bounds.

Perhaps something like
```Julia
mean = Interval(0, 1)
var = Interval(1, 2)
l = ProbabilityBox(Normal, [mean, std] , :l) # length
```

This is only a personal preference though.

test/reliability/probabilityoffailure.jl
AnderGray: It's a nice example. Should be added a demo.

src/reliability/probabilityoffailure.jl
```diff
@@ -1,92 +1,145 @@
...
+    lb, ub = bounds(inputs)
+    x0 = (lb .+ ub) ./ 2
+
+    rhobeg = minimum((ub .- lb) ./ 2)
+    _, info_min = prima(montecarlo_pf, x0; xl=lb, xu=ub, rhobeg=rhobeg)
+    pf_lb = info_min.fx
+    _, info_max = prima(x -> -montecarlo_pf(x), x0; xl=lb, xu=ub, rhobeg=rhobeg)
+    pf_ub = -info_max.fx
+
+    return Interval(pf_lb, pf_ub, :pf)
```
AnderGray, line 121: The function is very nice. 

We should also be able to return samples of the extremal distributions right? I think this would be the two distributions which maximise and minimise the pf.

The same could apply to the COV ... however this might be trickier ... it's at least known how to return a confidence interval for an imprecise Monte Carlo simulation.

AnderGray (commented):
Thanks for pushing this forward! I think as a first implementation it's looking good. 

Your imprecise failure probability function looks general enough to apply to methods other than MC :)

src/reliability/probabilityoffailure.jl
andreaperin: I have to create a "super type" for the abstract simulation, hopefully next week!

FriesischScott: Yeah, Andrea and I already discussed having a proper `AbstractSimulation` for these situations. We will then subtype `AbstractMonteCarlo` from it to separate.

src/inputs/imprecise/p-box.jl
FriesischScott: I think you are right, and I think @andreaperin wanted this as well. It's just a bit more tedious to collect the values for the upper/lower bounds but nothing major. And will be a lot more intuitive.

FriesischScott: Having only `UQInput` and `ImpreciseUQInput` won't work if you have both in your inputs. Unfortunately we need a supertype.

src/UncertaintyQuantification.jl
FriesischScott: Sadly, I think we need both.

src/reliability/probabilityoffailure.jl
FriesischScott: For the demo I think we should use an example from a paper somewhwere with a reference solution.

src/inputs/imprecise/p-box.jl
```diff
@@ -0,0 +1,38 @@
...
+# Examples
+
+```jldoctest
+julia>  ProbabilityBox([1.75, 1.83], [1.77, 1.85], x -> Uniform(x...), :l)
+ProbabilityBox(Real[1.75, 1.83], Real[1.77, 1.85], var"#7#8"(), :l)```
+"""
+struct ProbabilityBox <: ImpreciseUQInput
+    lb::AbstractVector{<:Real}
+    ub::AbstractVector{<:Real}
+    dist::Function
```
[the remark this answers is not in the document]
FriesischScott: I think I'd like to have this already in the first version. It'll be a bit annoying to write, but a lot nicer for the user.

src/inputs/imprecise/interval.jl
```diff
@@ -0,0 +1,34 @@
...
+
+function map_to_precise(x::Real, input::Interval)
+    lb = input.lb
+    ub = input.ub
+    x < lb && error("Choosen value $x is lower than Interval's lower bound $lb")
+    x > ub && error("Choosen value $x is higher than Interval's upper bound $ub")
+    return Parameter(x, input.name)
+end
+
+function sample(i::Interval)
```
FriesischScott, line 32: Maybe we should use a different name here., as it's not really sampling.

src/reliability/probabilityoffailure.jl
```diff
@@ -1,92 +1,196 @@
-function probability_of_failure(
```
FriesischScott, line 1: Maybe move the functions in a file `imprecise_probabilityoffailure.jl` or something like that. Might be worth keeping the precise/imprecise stuff separately.

src/reliability/probabilityoffailure.jl
FriesischScott: @AnderGray is right, it would be nice to return the parameters, rvs, etc leading to the upper and lower bounds. So you could for example run other analyses after the reliability analysis for the upper and lower bounds.

test/reliability/probabilityoffailure.jl
```diff
@@ -1,90 +1,130 @@
-@testset "Probability of Failure" begin
```
FriesischScott, line 1: If the code goes into a new file, so should the tests.

FriesischScott (changes requested):
See the comments for required changes. I'd also like to see the benchmarking for the three global optimization libraries. You can use `BenchmarkTools.jl`.

FriesischScott:
@AnderGray Thoughts on the second method with inner optimization?

AnderGray (commented):
Looking good. The 2nd imprecise method is very cool. Sometimes called 'random slicing', as it's based on random set theory.

What's implemented here I believe is interval Monte Carlo, and it's quite expensive. I don't know if we should tackle it in this PR, but this can be generalised to other reliability methods (FORM, importance sampling, linesampling and even SuS).

See paper by Diego Alvarez: https://doi.org/10.1016/j.ymssp.2017.07.040

The method is quite simple. You 'just' need to solve 2 imprecise reliability problems, for the upper (lower) performance functions:

Screenshot 2024-02-21 at 23 11 57

The idea is that each `u`, a sample in SNS, has an associated upper (lower) performance function, which is found using interval prop with optimisation. So for each evaluate SNS sample from FORM/LS/SuS, you to do an interval optimisation to find the upper bound on `G`. This can be used a standard performance function in a precise reliability algorithm. This is repeated for the lower bound.

You end up with this kind of picture (physical space, copula space, SNS space)

Screenshot 2024-02-21 at 23 54 16

For implementation, I think it's easiest to wrap the interval performance function optimisation (like you did for Monte Carlo), and pass it to the other reliability algorithms. 

We'd also need a function to go from SNS to for p-box. This isn't too challenging under independence.

src/inputs/imprecise/p-box.jl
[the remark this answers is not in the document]
FriesischScott: I solved this by using a parametric struct. It's quite easy and intuitive.

src/inputs/imprecise/p-box.jl
FriesischScott: While I really like specifying the parameters as intervals, here it creates issues because we already have an `Interval` type with the extra `name::Symbol`.

FriesischScott: Mhh, I believe we could actually do it like this:
```julia
μ = Interval(0, 1, :μ)
σ = Interval(0.1, 1, :σ)
l = ProbabilityBox{Normal}([μ, σ] , :l) # length
```

Internally, we can convert the parameters to a named tuple whenever needed and pass them to the keyword constructors of the distribution. It would actually eradicate the need to have the parameters in order as well. You'd just have to use the corresponding symbols for the keywords in *Distributions.jl*. Plus it looks so clean 😄. Great suggestion.

FriesischScott:
I think we should implement proper types to dispatch on instead of dispatching on the `inputs` vector like

- `DoubleLoop(MonteCarlo(10000))`
-  `IntervalMonteCarlo(1000)`

Each  method should than check, that at least one *imprecise* input is passed and error otherwise.

FriesischScott:
On that note, we might also define the input types with the proper terminology:

- `DeterministicUQInput`
- `AleatoryUQInput`
- `EpistemicUQInput`
- `HybridUQInput`

Is this too complex? The user will usually not have to interact with these types at all. They just create their `Parameter`, `RandomVariable`, `Interval` or `ProbabilityBox` and move on. But for anyone already familiar with the terms it could be nice.

FriesischScott:
> On that note, we might also define the input types with the proper terminology:
> 
>     * `DeterministicUQInput`
> 
>     * `AleatoryUQInput`
> 
>     * `EpistemicUQInput`
> 
>     * `HybridUQInput`
> 
> 
> Is this too complex? The user will usually not have to interact with these types at all. They just create their `Parameter`, `RandomVariable`, `Interval` or `ProbabilityBox` and move on. But for anyone already familiar with the terms it could be nice.

Yeah, this is too much and unnecessary. if we dispatch on the simulation all we need is a precise and imprecise type like we have now.

FriesischScott:
> > On that note, we might also define the input types with the proper terminology:
> > ```
> > * `DeterministicUQInput`
> > 
> > * `AleatoryUQInput`
> > 
> > * `EpistemicUQInput`
> > 
> > * `HybridUQInput`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is this too complex? The user will usually not have to interact with these types at all. They just create their `Parameter`, `RandomVariable`, `Interval` or `ProbabilityBox` and move on. But for anyone already familiar with the terms it could be nice.
> 
> Yeah, this is too much and unnecessary. if we dispatch on the simulation all we need is a precise and imprecise type like we have now.

@AnderGray  and I decided to just keep `UQInput` and `ImpreciseUQInput` we can filter for the *precise* inputs using `!isa`

FriesischScott (approved):

AnderGray (approved):

PR #149, opened 2023-12-23
Fix (improve) unstable tests

### LINKED ISSUE #148: Unstable tests
Some tests are a bit unstable and fail randomly. Having to always re-run the actions is annoying.

Unstable tests include:

- [ ] EmpiricalDistribution > Rayleigh
- [ ] RandomVariable > SNS Mappings

FriesischScott:
This should fix the unstable tests for empirical distributions and mappings to standard normal space.

Closes #148.


FriesischScott:
Something seems very wrong here. It fails the test for samples drawn from a normal. They aren't even transformed. How is the pvalue this low?

FriesischScott:
Are hypothesis test just useless? Even if I sample a standard normal, the test will return anything between 0 and 1.

FriesischScott:
While the hypothesis tests didn't work out, this at least improves the stability of the tests as far as I can tell.

Issue #153, opened 2024-02-01
Better HPC interfacing


AnderGray:
Julia's inbuilt parallelism isn't really good for launching complicated workflows / models on clusters. Would be useful to have something that can interface better to the scheduler.

Example: 
- some of my simulations need to be run on multiple nodes, rather than a single cpu
- a decent amount of data processing is carried out before the model is run, and I run into memory issues

@FriesischScott has suggested a nice trick, where the solver calls `sbatch` and waits for the job to complete:

```Julia
OpenSees = Solver(
    "sbatch",
    "launch_sim.sh";
    args="--wait",
)
```
with `launch_sim.sh` containing everything need to run my workflow.

Nice, but, has some problems:

- To run parallel jobs, I need still need to `addprocs(N)`, and these N processors are idly waiting for jobs to finish.
- My scheduler maxes at 20 jobs at the same time, so if I want to run many quick model evaluations, this is slow.
- Launching many jobs like this can quickly swamp the scheduler.

## Slurm job arrays

[Slurm's job arrays](https://slurm.schedmd.com/job_array.html) is a nice way to manage many similar jobs, like jobs which differ by just an index (e.g. sample-N). Also allows you to preallocate total amount of resources needed, and how many are concurrently run etc.

It would require a bit of engineering, but perhaps we could an interface of some sort could be written ... i.e. when pmap is called in External model, the input files are created (directories created, files copied, values interpolated), and a slurm array submitted which will loop through the individual jobs.

## Maestro workflow manager

Could also be something to look into:
https://github.com/LLNL/maestrowf?tab=readme-ov-file
https://maestrowf.readthedocs.io/en/latest/index.html

Provides a yaml and command line tool for performing parameter studies, and can work with slurm.

Scheduling studies: https://maestrowf.readthedocs.io/en/latest/Maestro/scheduling.html#flux

PR #155, opened 2024-02-07
Slurm job array interface

AnderGray:
Adds the functionality discussed in  #153

We can now pass a `SlurmInterface` type to external model, and any call to execute! will be submitted as a job array. 

```Julia
struct SlurmInterface
    name::String
    account::String
    partition::String
    nodes::Integer
    ntasks::Integer
    batchsize::Integer
    extras::Vector{String}
    time::String
end
```

Julia waits until job array competes. In a slurm environment, allows you also to run parallel sampling in parallel but with Julia running on a single node (e.g. from the login node). Submitted external model runs can now be as heavy as you like.

Also added
- slurm-opensees.jl example
- slurm-openmc.jl example 

Working, but some things to do:
- More constructors
- Docstrings
- Variable names check 
- Testing?

### DIFF
--- a/demo/reliability/slurm-openmc.jl
+++ b/demo/reliability/slurm-openmc.jl
@@ -7,7 +7,7 @@ using UncertaintyQuantification, DelimitedFiles
 # @everywhere using UncertaintyQuantification, DelimitedFiles
 
 E = RandomVariable(Normal(40, 60), :Enrich)
-O = RandomVariable(Uniform(530, 690), :OuterWALL)
+O = RandomVariable(Uniform(530, 690), :OuterWall)
 
 # Source/Extra files are expected to be in this folder
 sourcedir = joinpath(pwd() * "/../..", "demo/models")
--- a/demo/reliability/slurm-openmc.jl
+++ b/demo/reliability/slurm-openmc.jl
@@ -6,7 +6,7 @@ using UncertaintyQuantification, DelimitedFiles
 # addprocs(6; exeflags="--project")
 # @everywhere using UncertaintyQuantification, DelimitedFiles
 
-E = RandomVariable(Normal(40, 60), :Enrich)
+E = RandomVariable(Uniform(40, 60), :Enrich)
 O = RandomVariable(Uniform(530, 690), :OuterWall)
 
 # Source/Extra files are expected to be in this folder
--- a/demo/reliability/slurm-openmc.jl
+++ b/demo/reliability/slurm-openmc.jl
@@ -1,0 +1,62 @@
+# Reference: https://opensees.berkeley.edu/wiki/index.php/Simply_supported_beam_modeled_with_two_dimensional_solid_elements
+using UncertaintyQuantification, DelimitedFiles
+
+# To run the model distributed add the desired workers and load the required packages with @everywhere
+# using Distributed, Formatting
+# addprocs(6; exeflags="--project")
+# @everywhere using UncertaintyQuantification, DelimitedFiles
+
+E = RandomVariable(Normal(40, 60), :Enrich)
+O = RandomVariable(Uniform(530, 690), :OuterWALL)
+
+# Source/Extra files are expected to be in this folder
+sourcedir = joinpath(pwd() * "/../..", "demo/models")
+
+# These files will be rendere through Mustach.jl and have values injected
+sourcefile = "openmc_TBR.py"
+
+# Dictionary to map format Strings (Formatting.jl) to variables
+numberformats = Dict(:E => ".8e")
+
+# UQ will create subfolders in here to run the solver and store the results
+workdir = joinpath(pwd(), "openmc_TBR")
+
+# Read output file and compute maximum (absolute) displacement
+# An extractor is based the working directory for the current sample
+TBR = Extractor(base -> begin
+    file = joinpath(base, "openmc.out")
+    data = readdlm(file, ' ')
+
+    return data[1]
+end, :TBR)
+
+openmc = Solver(
+    "python3", # path to OpenSees binary
+    "openmc_TBR.py";
+    args="", # (optional) extra arguments passed to the solver
+)
+
+slurm = SlurmInterface(;
+    name="UQ_slurm",
+    account="UKAEA-AP001-CPU",
+    partition="cclake",
+    nodes=1,
+    ntasks=1,
+    batchsize=200,
+    time="00:01:00",
+    extras=["module load openmc", "source ~/.virtualenvs/openmc/bin/activate"],
+)
+
+ext = ExternalModel(
+    sourcedir,
+    sourcefile,
+    TBR,
+    openmc;
+    workdir=workdir,
+    formats=numberformats,
+    slurm=slurm,
+)
+
+pf, samples = probability_of_failure(ext, df -> 1.0 .- df.TBR, [E, O], MonteCarlo(1000))
+
+println("Probability of failure: $pf")
--- a/src/models/externalmodel.jl
+++ b/src/models/externalmodel.jl
@@ -98,11 +98,7 @@ function evaluate!(
     n = size(df, 1)
     digits = ndigits(n)
 
-    ## get current dir
-
     for i = 1:n
-
-        ## set current dir
 
         path = joinpath(m.workdir, datetime, "sample-$(lpad(i, digits, "0"))")
         mkpath(path)
@@ -125,9 +121,7 @@ function evaluate!(
         end
     end
 
-    make_input(slurm)
-    launch(slurm)
-
+    run_slurm_Array(slurm, m, n)
 
     results = map(1:n) do i
         path = joinpath(m.workdir, datetime, "sample-$(lpad(i, digits, "0"))")
--- a/src/models/externalmodel.jl
+++ b/src/models/externalmodel.jl
@@ -48,7 +48,8 @@ function evaluate!(
     df::DataFrame;
     datetime::String=Dates.format(now(), "YYYY-mm-dd-HH-MM-SS"),
 )
-    if !isnothing(m.slurm) return evaluate!(m, df, datetime, m.slurm) end
+    if !isnothing(m.slurm) return evaluate!(m, df, m.slurm, datetime = datetime) end
+
     n = size(df, 1)
     digits = ndigits(n)
 
@@ -91,9 +92,9 @@ end
 
 function evaluate!(
     m::ExternalModel,
-    df::DataFrame;
+    df::DataFrame,
+    slurm::SlurmInterface;
     datetime::String=Dates.format(now(), "YYYY-mm-dd-HH-MM-SS"),
-    slurm::SlurmInterface
 )
     n = size(df, 1)
     digits = ndigits(n)
@@ -121,7 +122,7 @@ function evaluate!(
         end
     end
 
-    run_slurm_Array(slurm, m, n)
+    run_slurm_array(slurm, m, n, datetime)
 
     results = map(1:n) do i
         path = joinpath(m.workdir, datetime, "sample-$(lpad(i, digits, "0"))")
--- a/src/models/externalmodel.jl
+++ b/src/models/externalmodel.jl
@@ -7,6 +7,7 @@ struct ExternalModel <: UQModel
     extras::Vector{String}
     formats::Dict{Symbol,String}
     cleanup::Bool
+    slurm::Union{SlurmInterface, Nothing}
 
     function ExternalModel(
         sourcedir::String,
@@ -17,10 +18,11 @@ struct ExternalModel <: UQModel
         extras::Union{String,Vector{String}},
         formats::Dict{Symbol,String},
         cleanup::Bool,
+        slurm::Union{SlurmInterface, Nothing}
     )
         sources, extractors, extras = wrap.([sources, extractors, extras])
         return new(
-            sourcedir, sources, extractors, solver, workdir, extras, formats, cleanup
+            sourcedir, sources, extractors, solver, workdir, extras, formats, cleanup, slurm
         )
     end
 end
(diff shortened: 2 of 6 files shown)


src/solvers/slurm.jl
```diff
@@ -0,0 +1,71 @@
...
+        write(file, "#SBATCH --nodes=$(SI.nodes)\n")
+        write(file, "#SBATCH --ntasks=$(SI.ntasks)\n")
+        write(file, "#SBATCH --time=$(SI.time)\n")
+        write(file, "#SBATCH --output=UncertaintyQuantification_%A-%a.out\n")
+        write(file, "#SBATCH --error=UncertaintyQuantification_%A-%a.err\n")
+        write(file, "#SBATCH --array=[1-$(n)]%$(SI.batchsize)\n")
+        write(file, "\n\n\n")
+        write(file, "#### EXTRAS ####\n")
+
+        for extra in extras
```
FriesischScott, line 56: Does including the files in the slurm script do anything? Or is it just for reference?

src/solvers/slurm.jl
```diff
@@ -0,0 +1,71 @@
...
+            write(file, "$extra\n")
+        end
+
+        write(file, "\n\n\n")
+        write(file, "#### RUN COMMAND ####\n")
+        write(file, "cd sample-\$(printf %0$(digits)d \$SLURM_ARRAY_TASK_ID)\n")
```
FriesischScott, line 62: This `cd` is local to SLURM and should not impact Julia, correct?

src/models/externalmodel.jl
```diff
@@ -85,6 +90,57 @@ function evaluate!(
...
+function evaluate!(
+    m::ExternalModel,
+    df::DataFrame,
+    slurm::SlurmInterface;
+    datetime::String=Dates.format(now(), "YYYY-mm-dd-HH-MM-SS"),
+)
+    n = size(df, 1)
+    digits = ndigits(n)
+
+    for i = 1:n
```
FriesischScott, line 102: This has a lot of code duplication with the original `evaluate!` method. Can you extract some of it into internal functions and use them in both?

FriesischScott (changes requested):
Overall very nice work! I haven't tried it yet but it looks clean and should get the job done. I included a few remarks.

FriesischScott:
For testing, some of the code could be extracted into internal functions. For example, one that writes the slurm script. This can then be tested, independently from actually running it.

src/solvers/slurm.jl
FriesischScott: Disregard, this is for loading modules.

demo/reliability/slurm-openmc.jl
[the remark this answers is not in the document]
AnderGray: Thanks for the review, and the suggestions 👍

src/solvers/slurm.jl
AnderGray: Yes correct. The generated `slurm_array.sh` is in the same directory as the `sample-XXX` directories, so at the beginning of each job, it enters a directory associated to `SLURM_ARRAY_TASK_ID`.

docs/literate/hpc/slurm-opensees.jl
FriesischScott, line 1: This can be removed. One example for Slurm is enough.

demo/hpc/slurm-opensees.jl
FriesischScott, line 1: Delete together with the source file.

src/hpc/slurm.jl
```diff
@@ -0,0 +1,167 @@
...
+    throttle  : the number of jobs to be run at the same time
+    jobname   : name of the job
+    mempercpu : string, amount of RAM to give per cpu (default MB)
+    extras    : instructions to be executed before the model is run, e.g. activating a python environment
+    time      : how long each job takes, e.g. 00:10:00 for 10 minutes per sample
+
+# Examples
+```jldoctest
+julia> slurm = SlurmInterface(account = "HPC_account_1", partition = "CPU_partition", nodes = 1, ntasks = 32, throttle = 200, extras = ["load python3"], time = "00:10:00")
+SlurmInterface("HPC_account_1", "CPU_partition", 1, 32, 200, "UQ_array", ["load python3"], "00:10:00")
```
FriesischScott, line 31: Looks like this doctest fails. Might be the empty lines.

docs/src/manual/hpc.md
```diff
@@ -0,0 +1,53 @@
...
+!!! note 
+    If your `extras` command requires `""` or `$` symbols, please use `\"\"` and `\$`.
+
+`throttle` specifies the job array task throttle: the number of samples that will be run at any given time. For example, when running a `MonteCarlo` simulation with 2000 samples, and `throttle = 50`, 2000 model evaluations will be run in total, but only 50 at the same time. If left empty, your scheduler's default throttle will be used.
+
+
+<!-- ## Useage
+
+!!! note
+    See [eamples/HPC](../examples/hpc.md) for a detailed example
```
FriesischScott, line 51: Typo. Also doesn't need to be a note.

docs/make.jl
```diff
@@ -23,8 +23,12 @@ makedocs(
             "Getting Started" => "manual/gettingstarted.md",
             "Reliability Analysis" => "manual/reliability.md",
             "Metamodelling" => "manual/metamodels.md",
+            "HPC" => "manual/hpc.md",
+        ],
+        "Examples" => [
+            "Metamodels" => "examples/metamodels.md", 
+            "HPC" => "examples/hpc.md",
```
FriesischScott, line 30: "High Performance Computing"

docs/src/manual/hpc.md
```diff
@@ -0,0 +1,53 @@
+# HPC
```
FriesischScott, line 1: `# High Performance Computing`

src/models/externalmodel.jl
```diff
@@ -7,6 +7,7 @@ struct ExternalModel <: UQModel
     extras::Vector{String}
     formats::Dict{Symbol,String}
     cleanup::Bool
+    slurm::Union{SlurmInterface,Nothing}
```
FriesischScott, line 10: We should call this `scheduler` to and use `AbstractHPCScheduler` to leave this extendible.

docs/src/manual/hpc.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## SlurmInterface
+
+When `SlurmInterface` is passed to an `ExternalModel`, a slurm job array script is automatically generated and executed. Julia waits for this job to finish before extracting results and proceeding.
+
+Your machine information must be passed to `SlurmInterface` for the array script to be correctly generated. For example:
+
+```@example hpc
+using UncertaintyQuantification
```
FriesischScott, line 16: This is clear and doesn't need to be repeated here.

docs/make.jl
```diff
@@ -23,8 +23,12 @@ makedocs(
             "Getting Started" => "manual/gettingstarted.md",
             "Reliability Analysis" => "manual/reliability.md",
             "Metamodelling" => "manual/metamodels.md",
+            "HPC" => "manual/hpc.md",
```
FriesischScott, line 26: "High Performance Computing"

docs/src/manual/hpc.md
```diff
@@ -0,0 +1,53 @@
...
+
+`extras` are commands that will be executed before you model is run, for example loading any module files or data your model requires. Multiple commands can be passed: `extras = ["load python", "python3 get_data.py"]`.
+
+!!! note 
+    If your `extras` command requires `""` or `$` symbols, please use `\"\"` and `\$`.
+
+`throttle` specifies the job array task throttle: the number of samples that will be run at any given time. For example, when running a `MonteCarlo` simulation with 2000 samples, and `throttle = 50`, 2000 model evaluations will be run in total, but only 50 at the same time. If left empty, your scheduler's default throttle will be used.
+
+
+<!-- ## Useage
```
FriesischScott, line 48: `### Usage`

docs/literate/hpc/slurm-openmc.jl
```diff
@@ -0,0 +1,106 @@
...
+)
+
+# Specify a limitstate function, negative value consititutes failure. Here we are interested in P(TBR <= 1).
+function limitstate(df)
+    return  reduce(vcat, df.TBR) .- 1 
+end
+
+
+# Run a Monte Carlo simulation
+@time pf, cov, samples = probability_of_failure(ext, limitstate, [E, R1], MonteCarlo(5000))
```
FriesischScott, line 78: I think by default Documenter will try to run this when building the documentation. Which won't work.

docs/literate/hpc/slurm-openmc.jl
```diff
@@ -0,0 +1,106 @@
...
+println("Probability of failure: $pf")
+
+
+#jl using StatsBase
+
+#jl TBR = samples.TBR #hide
```
FriesischScott, line 84: `# hide` is used to hide output in the docs. Not necessary for stuff that only goes into the Julia file.

docs/literate/hpc/slurm-openmc.jl
```diff
@@ -0,0 +1,106 @@
+#===
+# HPC
```
FriesischScott, line 2: `# High Performance Computing`

docs/literate/hpc/slurm-openmc.jl
```diff
@@ -0,0 +1,106 @@
...
+## OpenMC TBR uncertainty
+
+In this example, we will run OpenMC, to compute the tritium breeding ratio (TBR) uncertainty, by varying material and geometric properties. This example was taken from: https://github.com/fusion-energy/neutronics-workshop
```
FriesischScott, line 5: Should actually be a link a the end.

docs/literate/hpc/slurm-openmc.jl
```diff
@@ -0,0 +1,106 @@
...
+This is also a good time to declare the function that we are working with.
+===#
+
+#===
+Here we will vary the models Li6 enrichment, and radius of the tungsten layer
+===#
+using UncertaintyQuantification, DelimitedFiles
+
+E = RandomVariable(Uniform(0.3, 0.7), :E)
+R1 = RandomVariable(Uniform(8, 14), :R1)
```
FriesischScott, line 18: I think most of the output can be hidden here. We don't need it in the docs.

test/hpc/slurm.jl
FriesischScott, line 1: Tests need to be included in `runtests.jl`.

src/hpc/slurm.jl
```diff
@@ -0,0 +1,167 @@
...
+    run_command = !isempty(args) ? "$binary $args $source" : "$binary $source"
+    array_command = if iszero(SI.throttle)
+        "#SBATCH --array=[1-$(n)]\n"
+    else
+        "#SBATCH --array=[1-$(n)]%$(SI.throttle)\n"
+    end
+
+    digits = ndigits(n)
+
+    fname = "slurm_array.sh"
```
FriesischScott, line 97: If you extract this into a method it will be easier to test.

src/models/externalmodel.jl
```diff
@@ -34,48 +36,91 @@ function ExternalModel(
...
     cleanup::Bool=false,
+    slurm::Union{SlurmInterface,Nothing}=nothing,
 )
     return ExternalModel(
-        sourcedir, sources, extractors, solver, workdir, extras, formats, cleanup
+        sourcedir, sources, extractors, solver, workdir, extras, formats, cleanup, slurm
     )
 end
 
+function makedirectory(m::ExternalModel, sample, path::String)
```
FriesischScott, line 46: Needs a test.

src/models/externalmodel.jl
```diff
@@ -34,48 +36,91 @@ function ExternalModel(
...
+        end
+    end
+
+    for file in m.extras
+        cp(joinpath(m.sourcedir, file), joinpath(path, file))
+    end
+    return nothing
+end
+
+function getresult(m::ExternalModel, path::String)
```
FriesischScott, line 68: Needs a test.

test/hpc/slurm.jl
FriesischScott: But should only be included on Linux.

test/hpc/slurm.jl
```diff
@@ -0,0 +1,123 @@
...
+    open(joinpath(sourcedir, "extra.txt"), "w") do input
+        println(input, "This is an extra file")
+    end
+
+    x = RandomVariable(Uniform(0, 1), :x)
+    y = RandomVariable(Uniform(0, 1), :y)
+
+    df = sample([x, y], 5)
+
+    @testset "No Cleanup" begin
```
FriesischScott, line 55: These don't need to be tested again.

FriesischScott (changes requested):
One of the main issues will be the docs. By default Literate renders Julia code into `@example` blocks which Documenter tries to run. I'm not sure how we can change that for certain files and leave it intact for others.

src/hpc/slurm.jl
FriesischScott: It's the missing memory in the output.

AnderGray:
I think ready for another look.

As discussed offline, the `SlurmInterface` is tested on Linux, by importing and overwriting the `run_HPC_job` function, which calls a dummy script which loops over the requested samples.

We discussed that aliasing the `slurm` command in the environment was a better solution, however we can seem to alias commands using Julia's shell commands. [See here](https://discourse.julialang.org/t/aliases-not-recognized-in-julia-shell-mode/51571), and [here](https://discourse.julialang.org/t/julias-shell-mode-cant-recognize-user-defined-aliases-of-system-commands-on-linux/109724) for discussion. We could perhaps pre-set this using GitHubs CI, and only test slurm on GitHub?

Also the gaussian copula test seems to fail sometimes. We should perhaps consider migrating to Copulas.jl, although there are missing features (such as Rossenblatt transform).

PR #160, opened 2024-03-20
update slurm_extras

yyimingucl:
add slurm_extras arguments to the HPC SlurmInterface for accepting some undefined slurm arguments.


FriesischScott:
Thanks for the PR @yyimingucl .  What are the exact  arguments currently missing that you need?

I see a small risk of conflicting options with defining them like this.

FriesischScott:
Closed in favour of #189

PR #168, opened 2024-05-13
Add high dim example to docs

AnderGray:
Nearly working new doc page. Shows scalability of Subset algorithms with `pf`, `N_samples`, and `N_dimensions`.

Branch of `modified_mcmc`


AnderGray:
The new example likely takes too long to compile in the docs. Should we generate the plots locally, and upload the pngs?

FriesischScott:
As discussed elsewhere we will introduce a new section in the docs specifically for benchmarks and case studies. These will not be executed when Documenter builds the docs so the images must be created and committed together with the markdown files.

PR #180, opened 2024-07-23
Stochastic dynamics framework

FriesischScott:
This PR adds basic support for stochastic processes through the spectral representation method.  More stypes of stochastic processes such as K-L expansion will follow later.

The process is represented and sampled through the internal random variables. This allows us to use advanced MC methods among other things. For convenience, we include a `StochasticProcessModel` (working title) that a user can use to compute the timeseries and add it to the `DataFrame` for later models.

There will be more to come and  sufficient documentation before the merge.


FriesischScott:
The extra deps are just for testing.

FriesischScott:
We're trying to put together a proper example using a model of a 42 story building in *OpenSees*.

PR #184, opened 2024-08-12
Add batch support for SLURM job arrays

FriesischScott:
This PR adds a new `batchsize` option to the `SlurmInterface` with a default of `0`. If left at `0` everything stays as it is now. When a batch size is specified we keep the current setup but run `ceil(n / batchsize)` job arrays sequentially before collecting the complete results and continuing the analysis. This is done by adding a batch id as the last arguments to the functions that generate and execute `HPC` jobs.


FriesischScott:
I'm testing it on our cluster now to see if we can run 10000 samples in batches of 300 🤞.

AnderGray:
A very useful feature, if your scheduler limits the number of job submission.

I added a couple of warnings about having `/test/test_utilities` in path, which I believe could cause failures when testing locally.

I additionally added an option to test the package on HPC using actual slurm:

```Julia
julia --project
using Pkg
Pkg.test(;test_args=["HPC", "YOUR_ACCOUNT", "YOUR_PARTITION"])
```
if you have the package cloned locally. Or if you have it installed from the registry, I believe the following works:

```Julia
using Pkg
Pkg.test("UncertaintyQuantification"; test_args=["HPC", "YOUR_ACCOUNT", "YOUR_PARTITION"])
```

I also moved the work directory of the generated simulation to someplace local, as I was getting weird behaviour with temp directories.

AnderGray (approved):

PR #193, opened 2024-09-23
Remove Formatting.jl

teobros:
Substitute the deprecated package Formatting.jl with Format.jl. All the instances of Formatting in the code and in the external model have been removed from the code. The toml files have been modified as well.


Project.toml
```diff
@@ -36,7 +36,6 @@ Dierckx = "0.5"
 Distributions = "0.24, 0.25"
 FastGaussQuadrature = "0.4, 0.5, 1"
 FiniteDifferences = "0.12"
```
FriesischScott, line 38: Needs a compat entry for Format

demo/reliability/supported-beam-opensees.jl
```diff
@@ -2,7 +2,7 @@
 using UncertaintyQuantification, DelimitedFiles
 
 # To run the model distributed add the desired workers and load the required packages with @everywhere
-# using Distributed, Formatting
+# using Distributed, Format
```
FriesischScott, line 5: This wasn't required anymore for a while so you only need to run `using Distributed`

test/runtests.jl
```diff
@@ -1,6 +1,6 @@
 using DataFrames
 using Distributed
-using Formatting
+using Format
```
FriesischScott, line 3: This might not be required anymore

test/Project.toml
```diff
@@ -1,7 +1,7 @@
 [deps]
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
-Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
+Format = "1fa38f19-a742-5d3f-a2b9-30dd87b9d5f8"
```
FriesischScott, line 4: Probably not needed anymore

FriesischScott (changes requested):
Add the compat entry and possibly remove the dependency  for the tests then it's good to go.

FriesischScott (approved):
LGTM!

PR #202, opened 2024-09-25
Add Advanced Line Sampling

lukasfritsch:
Add advanced line sampling along with an example and tests


AnderGray:
Looks great! I've tried it with p-boxes, and it's working 🥇 (at least it runs). Giving me similar accuracy as LineSampling, but `500 vs 4000` model evaluations.

A couple of initial things after discussion:

* Probably advanced line sampling should go in the same file as line sampling?
* You've implemented a splinefit function, is this the same as that's used in standard line sampling. If so, can we re-use the function? Perhaps don't dispatch on `AdvancedLineSampling`, just passing in `sim.points`
* Perhaps change name `_grad` to `_grad1D`. Pass the step size instead of dispatching on simulation. 
* `evaluate!` needs to happen to run model. This is best in `probability_of_failure` for the first spline, and then in `_grad1D`. But performance still can be evaluated here if you wish, which is always a native Julia function.
* I'm finding that 10 iterations on the root finding often leads to no root being found. This is happening particularly with p-boxes.
* Algorithm is serial, is there any way to parallelise? No problem if not, it just means that standard line sample is perhaps more attractive is you have lots and lots of compute available. One way to perhaps do this is to use spline interpolation for all lines, however after discussion, we think most of the efficiency gain is in the newton rapson.

lukasfritsch:
Updated the code based on the comments from @AnderGray

demo/reliability/linesampling.jl
```diff
@@ -0,0 +1,44 @@
...
+# Perform Line Sampling
+LS = LineSampling(numberlines, collect(0.5:0.5:10))
+pf_ls, σ_pf_ls, samples_ls = probability_of_failure([y], g, x, LS)
+cov_ls = σ_pf_ls/pf_ls
+
+# Perform Advanced Line Sampling
+ALS = AdvancedLineSampling(numberlines, collect(0.5:0.5:10))
+pf_als, σ_pf_als, samples_als = probability_of_failure([y], g, x, ALS)
+cov_als = σ_pf_als/pf_als
+
```
AnderGray, line 35: Add a Monte Carlo simulation to check results:

```Julia
pf_MC, σ_pf_MC, samples_MC = probability_of_failure([y], g, x, LatinHypercubeSampling(10_000_000))
cov_MC = σ_pf_MC/pf_MC
```

demo/reliability/linesampling.jl
```diff
@@ -0,0 +1,44 @@
...
+println("Line Sampling:")
+println("Pf = $pf_ls")
+println("CoV = $cov_ls")
+println("Model Calls = $(length(samples_ls.x1))\n")
+
+println("Advanced Line Sampling:")
+println("Pf = $pf_als")
+println("CoV = $cov_als")
+println("Model Calls = $(length(samples_als.x1))")
```
AnderGray, line 44: Show Monte Carlo results

```Julia
println("Monte Carlo:")
println("Pf = $pf_MC")
println("CoV = $cov_MC")
println("Model Calls = $(length(samples_MC.x1)) \n")
```

AnderGray, line 44: Add an imprecise example. I've commented out the `RandomSlicing` example, due to issue #211

```Julia
println("Performing Imprecise Reliability Analysis")

mean = Interval(-0.2, 0.2, :μ)
std = Interval(0.8, 1.2, :σ)
x_pbox = ProbabilityBox{Normal}.(Ref([mean, std]), [:x1, :x2])


# Perform Advanced Line Sampling
@time pf_als_DL = probability_of_failure([y], g, x_pbox, DoubleLoop(AdvancedLineSampling(numberlines, collect(0.5:0.5:10))))
# @time pf_als_RS, σ_pf_als_RS, samples_als_RS = probability_of_failure([y], g, x_pbox, RandomSlicing(AdvancedLineSampling(numberlines, collect(0.5:0.5:10))))

@time pf_ls_DL = probability_of_failure([y], g, x_pbox, DoubleLoop(LineSampling(numberlines, collect(0.5:0.5:10))))
# @time pf_ls_RS, σ_pf_ls_RS, samples_ls_RS = probability_of_failure([y], g, x_pbox, RandomSlicing(LineSampling(numberlines, collect(0.5:0.5:10))))


println("Double Loop:")
println("Line Sample pf = [$(pf_ls_DL.lb), $(pf_ls_DL.ub)]")
println("Advanced Line Sample pf = [$(pf_als_DL.lb), $(pf_als_DL.ub)] \n")

# println("Random Slicing:")
# println("Line Sample pf = $pf_ls_RS")
# println("Advanced Line Sample pf = $pf_als_RS")

```

AnderGray (changes requested):
I think we're nearly there.

Apart from the minor requests, a final additional thing to do is to make sure that your added files are formatted correctly. You can do this using the [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl).

Please add it outside of the package environment, then format the code with:

```Julia
using JuliaFormatter
format("src/linesampling.jl")
```

Can you make sure to run this on all your added files.

lukasfritsch:
Implemented requested changes from @AnderGray 

Further added handling of the case that the iterative root finding returns `Inf` when the maximum iteration steps are reached. In that case a root finding using `rootinterpolation()` with the specified points along the line is performed.

AnderGray (approved):
All great. Thanks for this very nice PR

PR #203, opened 2024-09-25
Variational inference issue#172

jgrashorn:
Added variational inference in the form of maximum likelihood and maximum a posteriori estimates.
Also wrote some tests and added examples in the bayesianupdating demo.


demo/bayesianupdating/inverse-eigenvalue.jl
FriesischScott, line 1: This file shouldn't be edited manually. You edit the file in `docs/literate/bayesianupdating` run `literateDocs.jl` to update the demo file.

Project.toml
FriesischScott, line 1: You need to also add a compat entry for Optim

src/UncertaintyQuantification.jl
```diff
@@ -96,6 +97,8 @@ export LeastSquares
 export LegendreBasis
 export LineSampling
 export SingleComponentMetropolisHastings
+export MaximumAPosteriori
```
FriesischScott, line 100: These should have some mention of variational inference in the name. Otherwise the user wouldn't know that something like `MaximumLikelihood` refers to  variational inference

src/modelupdating/bayesianupdating.jl
```diff
@@ -320,3 +320,129 @@ function _beta_and_weights(β::Real, L::AbstractVector{<:Real})
...
+"""
+
+"""
+Maximum a posteriori estimate, uses optimization to find the most likely model parameters based on prior and likelihood
+"""
+## !TODO implement abstract variational inference for ez change of method?
+struct MaximumAPosteriori <: AbstractBayesianMethod # Gaussian approximation with variational inference
+    # ! TODO: find out how to optimize
+    prior::Vector{RandomVariable}
+    optimMethod::Any
```
FriesischScott, line 336: Don't use `Any` here.

Does the user even have to select the method or is it fine to hardcode it?

src/modelupdating/bayesianupdating.jl
```diff
@@ -320,3 +320,129 @@ function _beta_and_weights(β::Real, L::AbstractVector{<:Real})
...
+
+"""
+Maximum likelihood estimate, uses optimization to find the maximum of the likelihood
+"""
+## !TODO Currently the prior is used to get information about model parameters, maybe there is a better way. In MLE the prior is not needed
+
+struct MaximumLikelihood <: AbstractBayesianMethod # Gaussian approximation with variational inference
+    # ! TODO: find out how to optimize
+    prior::Vector{RandomVariable}
+    optimMethod::Any
```
FriesischScott, line 397: see above

test/modelupdating/bayesianupdating.jl
```diff
@@ -251,4 +251,48 @@
...
+
+        estimater = MaximumAPosteriori([prior_sample_], optMethod, x0)
+
+        MAPEst, analytic_mean, analytic_std = normalmeanbenchmark(estimater, prior)
+
+        @test mean(MAPEst[1]) ≈ analytic_mean rtol = 0.05
+
+    end
+
+    @testset "MAP normal mean analytical" begin
```
FriesischScott, line 277: Looks like this  tests the maximum likelihood version

FriesischScott (changes requested):
I have a few comments and your tests don't seem to work.

src/UncertaintyQuantification.jl
jgrashorn: Wouldn't that be the same problem for SingleComponentMH and TMCMC?

src/modelupdating/bayesianupdating.jl
jgrashorn: see above

test/modelupdating/bayesianupdating.jl
jgrashorn: true

jgrashorn:
Comments have been implemented. Also changed the approximations to be able to handle multi-modal distributions when the user supplies more than one initial point.
Both methods have been added to the existing model updating example.

FriesischScott (changes requested):
A few more comments.

We need some documentation beyond just the inverse eigenvalue example.

src/modelupdating/bayesianupdating.jl
```diff
@@ -320,3 +320,173 @@ function _beta_and_weights(β::Real, L::AbstractVector{<:Real})
...
+    bayesianupdating(prior, likelihood, models, approximation)
+
+end
+
+"""
+Maximum likelihood estimate
+"""
+## !TODO Currently the prior is used to get information about model parameters, maybe there is a better way. In MLE the prior is not needed
+
+struct MaximumLikelihood <: AbstractBayesianMethod
```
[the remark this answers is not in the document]
jgrashorn: MAP and MLE technically do not qualify as variational inference in most cases, I might have mixed something up in the naming. Would for example `MaximumLikelihoodBayesian` or `MaximumLikelihoodBayesianUpdating` be sufficient?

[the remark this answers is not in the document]
FriesischScott: Now I'm, confused.  is this or is it not variational inference?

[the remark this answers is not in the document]
FriesischScott: And if not, what is this?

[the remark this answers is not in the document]
jgrashorn: It's not, but they can be used as basis for VI. Sorry for the confusion!

jgrashorn:
Added file for point estimates, renamed all methods with a -Bayesian suffix for easier identification. Moved the optimization and the setup into a function taking an `AbstractBayesianPointEstimate` argument.
Implemented the existing binomial tests for the point estimation methods. Added (optional) box constraints to optimization since they are sometimes needed for bounded variables (e.g. in Beta distribution).
Documentation is still needed.

jgrashorn:
Wrote some documentation, might still be lacking, but there are now one example each in `docs/literate/bayesianupdating` and in the manual, with some explanation.

docs/src/manual/bayesianupdating.md
```diff
@@ -220,6 +220,63 @@ tmcmc = TransitionalMarkovChainMonteCarlo(RandomVariable.(Uniform(-2,2), [:x, :y
...
+sample_points = reduce(vcat,[[x y] for x in xs, y in ys][:]) # hide
+df_points = DataFrame(sample_points, :auto) # hide
+likelihood_eval = exp.(loglikelihood(df_points)) # hide
+prior_eval = priorFunction(df_points) # hide
+contour!(xs, ys, likelihood_eval, lim = [-2,2], legend = :none) # hide
+contour!(xs, ys, prior_eval, lim = [-2,2], legend = :none) # hide
+savefig("point-estimates.svg"); nothing # hise
+```
+
+![Point estimates](point-estimates.svg)
```
AnderGray, line 278: Can you double check that the figure is correct? Here's what it looks like when I compile the docs.

image

src/modelupdating/bayesianpointestimate.jl
```diff
@@ -0,0 +1,176 @@
+"""
+Bayesian Point estimates
+"""
+
+"""
+Maximum a posteriori and maximum likelihood estimates, uses optimization to find the most likely model parameters based on prior and/or likelihood.
```
AnderGray, line 6: Maybe annotate the docstrings about more. What are the input arguments to the constructor do?

src/modelupdating/bayesianpointestimate.jl
```diff
@@ -0,0 +1,176 @@
...
+        lowerbounds::Vector{Float64} = [-Inf],
+        upperbounds::Vector{Float64} = [Inf]
+    )
+        return new(prior, optimmethod, x0, islog, lowerbounds, upperbounds)
+    end
+
+end
+
+"""
+general function for bayesianupdating for point estimate methods. Function takes likelihood, models, estimation method and an optional prior distribution. For MLE, the prior is ignored, if no prior is given, the prior evaluation function is generated from the given prior in the struct. For both methods, the variables to be updated are inferred from the given prior in the struct, so even if the prior is not needed in MLE, it should be given as a dummy-distribution to inform the method on which variables to update.
```
AnderGray, line 82: Can you write this is the same way as is written for the mcmc methods? As in the docstrings here:

https://github.com/FriesischScott/UncertaintyQuantification.jl/blob/bfdbc528452ede2930f0e167f7644830a8ca10dd/src/modelupdating/bayesianupdating.jl#L35

src/modelupdating/bayesianpointestimate.jl
```diff
@@ -0,0 +1,176 @@
...
+function optimize_pointestimate(optimTarget::Function, pointestimate::AbstractBayesianPointEstimate)
+
+    method = ""
+
+    if pointestimate.optimmethod=="LBFGS"
+        method = LBFGS()
+    elseif pointestimate.optimmethod=="NelderMead"
+        method = NelderMead()
+    else
+        error("Optimization method $(pointestimate.optimmethod) is not supported in UncertaintyQuantification.jl")
```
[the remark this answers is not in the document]
jgrashorn: There is no specific reason. As far as I know, LBFGS and NelderMead cover the most usecases.
We decided to implement it this way so that the user does not have to import Optim in order to use these estimates, but then we also have to manually change the methods that are implemented. It could be an idea (either in this PR or in the future) to change the constructors such that the user can either input a string or directly a method from Optim.

test/modelupdating/bayesianpointestimate.jl
```diff
@@ -0,0 +1,185 @@
...
+
+    @testset "MAP binomial test" begin
+
+        optMethod = "LBFGS"
+        x0 = [0.5]
+        
+        prior_Function = Beta(1,1)
+        prior = RandomVariable(prior_Function, :x)
+
+        estimater = MaximumAPosterioriBayesian([prior], optMethod, x0; lowerbounds = [0.], upperbounds = [1.])
```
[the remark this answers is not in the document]
jgrashorn: Good catch. I changed the test to compare results of MAP and MLE to the mode of the beta distribution.
In this test-case, since Beta(1,1) is a uniform distribution, MAP and MLE are the same.

docs/src/manual/bayesianupdating.md
jgrashorn: I have added legend entries to the estimates and a small description to the figure in the text. It seems to not be possible to add legend entries for contour plots in Julia?
The figure should be clearer now.

plot_1

src/modelupdating/bayesianpointestimate.jl
```diff
@@ -0,0 +1,176 @@
...
+        target(input)[1]*-1
+    end
+
+    return optimTarget
+
+end
+
+function optimize_pointestimate(optimTarget::Function, pointestimate::AbstractBayesianPointEstimate)
+
+    method = ""
```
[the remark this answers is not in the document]
jgrashorn: changed

src/modelupdating/bayesianpointestimate.jl
jgrashorn: Added the annotations for the constructors to be similar to the mcmc methods

src/modelupdating/bayesianpointestimate.jl
jgrashorn: Added this for both point estimation methods and also added more context to the explanations.

jgrashorn:
> I also find the naming "point estimates" slightly odd. I guess you wanted to generalised "maximum likelihood estimation" and "MAP"?

I am also not happy about the naming, but couldn't come up with a better name to summarize what MLE and MAP are doing.

docs/src/manual/bayesianupdating.md
AnderGray: Thanks, looks better. I guess the red is the posterior, and the blue is the prior. Is there a reason that the MAP point isn't at the maximum of the posterior?

src/modelupdating/bayesianpointestimate.jl
AnderGray: Thank you

AnderGray:
> I am also not happy about the naming, but couldn't come up with a better name to summarize what MLE and MAP are doing.

Thanks for the fixes. Maybe we could just pick one, either max likelihood or MAP, and use it for naming the files/titles. I swing towards MAP cause then it's a Bayesian method. It should be quite clear that if we can do one, we can do the other. Maybe better understood than "Bayesian point estimates".

In the docstrings, you can also link stuff, so people can find them:

I.e. at the end of the MAP docstrings add

```Julia
See also [`MaximumLikelihoodBayesian`](@ref), [`bayesianupdating `](@ref),  [`TransitionalMarkovChainMonteCarlo`](@ref).
```

And at the end of max likelihood docstrings add:

```Julia
See also [`MaximumAPosterioriBayesian`](@ref), [`bayesianupdating `](@ref),  [`TransitionalMarkovChainMonteCarlo`](@ref).
```

AnderGray:
Thanks for the changes, good from my side 👍

FriesischScott (approved):

PR #206, opened 2024-09-25
Add evaluation of support for ProbabilityBoxes with Uniform

### LINKED ISSUE #204: Incorrect Lower and Upper Bounds for P-Box with Uniform Distribution
When creating a ```ProbabilityBox``` from ```Uniform``` distribution without specifying ```lb``` and ```up``` they default to 0 and 1, respectively.

andreaperin:
closes #204

### DIFF
--- a/src/inputs/imprecise/p-box.jl
+++ b/src/inputs/imprecise/p-box.jl
@@ -48,8 +48,17 @@ end
 function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
-    domain = support(T())
-    return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    if isa(T(), Uniform)
+        v_ip = mapreduce(
+            x -> collect(bounds(x)), vcat, filter(x -> isa(x, ImpreciseUQInput), p)
+        )
+        v_p = map(x -> x.value, filter(x -> isa(x, Parameter), p))
+        values = vcat(v_ip, v_p)
+        return ProbabilityBox{T}(p, name, minimum(values), maximum(values))
+    else
+        domain = support(T())
+        return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    end
 end
 
 function ProbabilityBox{T}(
--- a/src/inputs/imprecise/p-box.jl
+++ b/src/inputs/imprecise/p-box.jl
@@ -49,11 +49,11 @@ function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
     if isa(T(), Uniform)
-        v_ip = mapreduce(
-            x -> collect(bounds(x)), vcat, filter(x -> isa(x, ImpreciseUQInput), p)
-        )
-        v_p = map(x -> x.value, filter(x -> isa(x, Parameter), p))
-        values = vcat(v_ip, v_p)
+        bounds_intervals = mapreduce(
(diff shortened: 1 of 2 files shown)


src/inputs/imprecise/p-box.jl
```diff
@@ -48,8 +48,17 @@ end
 function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
-    domain = support(T())
-    return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    if isa(T(), Uniform)
```
FriesischScott, line 51: I think this can be `T` because that's already the type.

src/inputs/imprecise/p-box.jl
```diff
@@ -48,8 +48,17 @@ end
 function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
-    domain = support(T())
-    return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    if isa(T(), Uniform)
+        v_ip = mapreduce(
+            x -> collect(bounds(x)), vcat, filter(x -> isa(x, ImpreciseUQInput), p)
```
FriesischScott, line 53: there is an `isimprecise` method you can use to filter

src/inputs/imprecise/p-box.jl
```diff
@@ -48,8 +48,17 @@ end
 function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
-    domain = support(T())
-    return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    if isa(T(), Uniform)
+        v_ip = mapreduce(
+            x -> collect(bounds(x)), vcat, filter(x -> isa(x, ImpreciseUQInput), p)
+        )
+        v_p = map(x -> x.value, filter(x -> isa(x, Parameter), p))
```
FriesischScott, line 55: The variable names could be a bit more telling and some comments  what you are actually doing would be nice.

FriesischScott (changes requested):
Good stuff, only minor comments

src/inputs/imprecise/p-box.jl
```diff
@@ -48,8 +48,16 @@ end
 function ProbabilityBox{T}(
     p::AbstractVector{<:UQInput}, name::Symbol
 ) where {T<:UnivariateDistribution}
-    domain = support(T())
-    return ProbabilityBox{T}(p, name, domain.lb, domain.ub)
+    # p-boxes with Uniform distribution as parameter must be treated separately since their support changes with p-box lower and upper bounds.
+    if isa(T(), Uniform)
```
FriesischScott, line 52: Turns out you can just do `T == Uniform`.

FriesischScott (changes requested):
Apart from this I think it's good to go. Although I am not entirely sure if we should allow truncating uniform pboxes at all.

src/inputs/imprecise/p-box.jl
AnderGray: I kind of agree about the truncating the uniform distribution. A truncated uniform is just another uniform. But, it might be different for a p-box?

If the application is for discretising a p-box into probability intervals, I believe there's a better way to do by directly using the upper and lower cdfs.

andreaperin: truncating a Uniform distribution just return a Uniform distributon with new parameter and not a "truncated" type

AnderGray (approved):

FriesischScott (approved):

PR #207, opened 2024-09-25
Pass slurm options as dict

FriesischScott:
This should give users a lot more flexibility when using slurm.

### DIFF
--- a/src/hpc/slurm.jl
+++ b/src/hpc/slurm.jl
@@ -105,8 +105,9 @@ function setup_slurm_array(
         write(file, "#!/bin/bash -l\n")
 
         for (k, v) in si.options
-            write(file, "SBATCH --$k=$v\n")
+            write(file, "#SBATCH --$k=$v\n")
         end
+
         write(
             file, "#SBATCH --output=sample-%$(digits)a/UncertaintyQuantification-%a.out\n"
         )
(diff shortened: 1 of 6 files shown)


src/hpc/slurm.jl
```diff
@@ -10,35 +10,41 @@ The `throttle` specifies the number of simulations in the job array which are ru
 
 # parameters
 
-    account   : the account to charge the computing time to
-    partition : the partition to use
-    nodes     : number of nodes per job
-    ntasks    : total number of cores
+    options   : SBATCH options to add to the slurm script
```
AnderGray, line 13: I think we should elaborate slightly in the doc strings.
 

`you can use batchsize to split the work into smaller batches.` 
to
`. In the case that your HPC machine limits the number of submitted jobs, you can split the submissions into smaller "batches". Specify "batchsize" to the maximum number of allowed simultaneously submissions. This does not change the total number of runs.`


and

`options   : SBATCH options to add to the slurm script` => `options   : A dictionary of SBATCH options to add to the slurm script`

and

`batchsize : maximum size of the slurm array` => `batchsize : maximum size of the slurm array, use when HPC limits the number of simultaneously submitted jobs`

**Also look at the `SlurmInterface` type definition at the top of the docstring**  it needs to be changed to match the new definition.

AnderGray (changes requested):
Some small changes, particularly related to docs and tests

AnderGray (approved):
Tests successfully ran on my HPC machine

PR #212, opened 2024-09-26
Fix design of experiment bounds

### LINKED ISSUE #205: Design of experiment for unbounded distributions returns wrong points
When inputs contain unbounded distributions and generating DOE points, the mapping from [-1, 1] to the phhysical value of the rvs put the point at -\sigma and +\sigma ignoring the value of the mean

teobros:
I have modified the doe test to catch the issue, since the old test was passing with the bug present. That happened because the test was using a zero mean normal to generate the doe points.
Then, the bug has been fixed by adding the mean to the bounds function.
Closes #205

### DIFF
--- a/src/simulations/doe.jl
+++ b/src/simulations/doe.jl
@@ -63,10 +63,10 @@ end
 
 function bounds(r::RandomVariable, σ::Int)
     lb = minimum(r)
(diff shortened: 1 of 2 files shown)


FriesischScott (approved):
LGTM!

PR #218, opened 2024-09-26
Add stochastic dynamics

FriesischScott:


src/UncertaintyQuantification.jl
```diff
@@ -121,15 +127,18 @@ export TwoLevelFactorial
 export bayesianupdating
 export calc
 export count_rvs
+export dft
```
AnderGray, line 130: Not sure if we should export everything. Do you think the user will be using our `dft` or `idft`? Perhaps they will which to use the FFT Julia package in their application, which could conflict with our export.

Do you need to export t `dft` and `idft` to use your features?

Feel free to disagree if this is incorrect

docs/src/manual/dynamics.md
```diff
@@ -0,0 +1,81 @@
+# Stochastic Dynamics
+Stochastic dynamics studies the behaviour of buildings and systems and their changes over time when they are affected by random processes, or also called stochastic processes, such as earthquakes or wind loads. Stochastic processes are characterised by an inherent randomness which can not be described deterministically[li2009sdos](@cite). To study the frequency distribution and their amplitude of a stochastic process, the so-called power spectral density (PSD) function can be defined.
```
AnderGray, line 2: You're missing a space between the references and the text. Also present in other locations throughout the doc

docs/src/manual/dynamics.md
```diff
@@ -0,0 +1,81 @@
...
+```math
+    x(t) = \sqrt{2} \sum \limits_{n=0}^{N_{\omega}-1}\sqrt{2 S(\omega_n) \Delta \omega} \cos(\omega_n t + \varphi),
+```
+where $\Delta \omega$ is the frequency increment, $\omega_n$ is the frequency at coordinate $n$, $t$ is the time vector and $N_{\omega}$ is the total number of frequency points.
+
+Stochastic processes generated by SRM can be utilised in a Monte Carlo framework, for instance, in order to determine the structural reliability or other quantities of interest.
+
+## PSD estimation
+If a time signal, such as an earthquake, is given and the time domain and shall be studied for their characteristics in frequency domain, the PSD estimation can be carried out. Although rigorous mathematical relationships between stochastic processes and the PSD function exist, estimation techniques are often necessary since exact determination would require continuous signals of infinite length, which are not practical in real-world applications.
```
AnderGray, line 21: Grammar mistake in `, is given and the time domain and shall be studied ...`. Maybe you have one `and` too many?

AnderGray, line 21: "...since `an` exact determination..."

demo/dynamics/single-degree-of-freedom.jl
```diff
@@ -0,0 +1,66 @@
...
+    end
+end
+
+displacement = Model(sdof, :d)
+
+inputs = [gm, m, k, c]
+
+models = [gm_model, displacement]
+
+# samples = sample(inputs, 10)
```
AnderGray, line 44: Delete commented code

src/dynamics/psd.jl
```diff
@@ -0,0 +1,51 @@
+
+abstract type AbstractPowerSpectralDensity end
+
```
AnderGray, line 3: Can you add some doc strings? I'm unfamiliar with these models, and do not know what the parameters's mean, or how to use these models.

test/inputs/stochasticprocesses/spectralrepresentation.jl
```diff
@@ -0,0 +1,41 @@
...
+    x = evaluate(srm_obj, ϕ[1, :])
+
+    Ŝ = Δt^2 / T0 * abs.(dft(x)) .^ 2 / (2π)
+    S̃ = periodogram(x,t)
+    S̄ = periodogram(x,t,false)
+
+    x_2 = srm_obj(collect(ϕ[1, :]))
+    x_3 = srm_obj(ϕ[1, :])
+
+    to_standard_normal_space!(srm_obj, ϕ)
```
AnderGray, line 26: Can you quickly describe what this testset is doing?

Also is `smr_obj` used / tested after the SNS transformation?

AnderGray (changes requested):
Thank you for this very nice PR!

I'm not really an expert on the stochastic process models / spectral representation, so I've only been able to comment on what I can understand.

I think most of the changes are quite light. The biggest one being if you could add some doc-strings in some places. I currently am unable to use these feature since I don't know what the parameter's mean.

Also, are all these stochastic processes Gaussian distributed? Or something else? What would I do if I had a covariance matrix/function and I wanted to generate samples of a gaussian random field. Could I use one of your models?

Also do you know of any examples which would have an analytical solution? I think this would make a good test for Monte Carlo and subset sim. Perhaps a simple sum of the process has known solution?

docs/src/manual/dynamics.md
marcobehrendt: Your're right, the `and` shall be `in`.

docs/src/manual/dynamics.md
marcobehrendt: It's fixed.

docs/src/manual/dynamics.md
marcobehrendt: Fixed.

src/UncertaintyQuantification.jl
marcobehrendt: We discussed this before implementation and felt that it would be better to implement our own version of dft and idft as other packages bring many dependencies. Our dft implementation will be used directly in the PSD estimation. But we can of course discuss this matter.

test/inputs/stochasticprocesses/spectralrepresentation.jl
marcobehrendt: The entire testset checks the generation of stochastic processes and the re-transformation into the frequency domain with our implemented methods. A benchmark can be found in the work of Shinozuka & Deodatis (1991), on which the test set is based. 
In the current version, `srm_obj` is not tested directly after the transformation to the standard normal space and back.

src/UncertaintyQuantification.jl
AnderGray: Yes, having our own implementation is a good idea. But do we need to export it, or is it only used internally? I.e. would the user ever have to call an `dft` to use your features? 

I ask this, because it can cause conflicts with other packages exporting ffts. However, after checking, I believe in the other package it's called `fft` vs our `dft`. So there wouldn't be a conflict anyway (at least with FFTW.jl).

test/inputs/stochasticprocesses/spectralrepresentation.jl
AnderGray: Okey, that works 👍

AnderGray (approved):
This is awesome 🔥 I have some minor questions, but not required for merging.

- Why is it called `StochasticProcessModel`. I feel like they're more `inputs` than `Models`
- Are we able to extract stochastic process parameters, or just input them into models?
- I'm not sure how to use your `SpectralRepresentation` model. This is for modelling earthquakes? I usually work with Gaussian stochastic processes, which I define using a covariance function/kernel, or a matrix when descritised. Maybe later this could be added.
- Is it feasible to work for imprecise inputs? Marginals and/or correlation parameters?

PR #223, opened 2024-09-30
Fix interval propagation for degenerate intervals

### LINKED ISSUE #211: Incorrect behaviour with Line Sampling and Random Slicing
```Julia

using UncertaintyQuantification

# Input
x = RandomVariable.(Normal(), [:x1, :x2])

# Model
y = Model(df -> df.x2 .+ 0.01 * df.x1 .^3.0 .+ sin.(df.x1), :y)

# Limit-state function
g(df) = 5 .- df.y

# Simulation
numberlines = 200
LS = LineSampling(numberlines, collect(0.5:0.5:10))

# Inputs
mean = Interval(-0.2, 0.2, :μ)
std = Parameter(1, :σ)
x_pbox = ProbabilityBox{Normal}.(Ref([mean, std]), [:x1, :x2])

# ERROR: MethodError: no method matching iterate(::UncertaintyQuantification.SlicingModel)
probability_of_failure([y], g, x_pbox, RandomSlicing(LS))
```

I have also observed a strange behaviour that when you perform a LineSampling simulation prior, and pass it to `RandomSlicing`, it seems that it tries passing `Intervals` to the model when evaluating instead of optimising `g`. I think `LS` is mutated in the previous simulation, and is leading to this behaviour:

```Julia
using UncertaintyQuantification

# Input
x = RandomVariable.(Normal(), [:x1, :x2])

# Model
y = Model(df -> df.x2 .+ 0.01 * df.x1 .^3.0 .+ sin.(df.x1), :y)

# Limit-state function
g(df) = 5 .- df.y

# Simulation
numberlines = 200
LS = LineSampling(numberlines, collect(0.5:0.5:10))

# Inputs
x = RandomVariable.(Normal(), [:x1, :x2])

mean = Interval(-0.2, 0.2, :μ)
std = Parameter(1, :σ)
x_pbox = ProbabilityBox{Normal}.(Ref([mean, std]), [:x1, :x2])

# Works
probability_of_failure([y], g, x, LS)

# ERROR: MethodError: no method matching ^(::Interval, ::Float64)
probability_of_failure([y], g, x_pbox, RandomSlicing(LS))
```

FriesischScott:
Pboxes now always return intervals even if the bounds are equal.

In the propagation, I am checking for degenerate intervals to treat them as precise values. There might be a very rare edge case here, where all intervals are degraded and none are left to propagate. We'll get to that once we encounter this issue.

Closes #211.

### DIFF
--- a/src/simulations/linesampling.jl
+++ b/src/simulations/linesampling.jl
@@ -9,7 +9,7 @@ mutable struct LineSampling <: AbstractSimulation
         direction::NamedTuple=NamedTuple(),
     )
 
-    if maximum(points) > 8.0
+    if maximum(points) > 8.12
         error("LineSampling does not support lines longer than 8.12.")
     end
         return new(lines, points, direction)
--- a/src/simulations/linesampling.jl
+++ b/src/simulations/linesampling.jl
@@ -8,6 +8,10 @@ mutable struct LineSampling <: AbstractSimulation
         points::Vector{<:Real}=collect(0.5:0.5:5),
         direction::NamedTuple=NamedTuple(),
     )
+
+    if maximum(points) > 8.0
+        error("LineSampling does not support lines longer than 8.12.")
+    end
         return new(lines, points, direction)
     end
 end
--- a/demo/reliability/linesampling.jl
+++ b/demo/reliability/linesampling.jl
@@ -18,18 +18,18 @@ x = RandomVariable.(Normal(), [:x1, :x2])
 y = Model(df -> df.x2 .+ 0.01 * df.x1 .^ 3 .+ sin.(df.x1), :y)
 
 # Limit-state function
-g(df) = 5 .- df.y
+g(df) = 3 .- df.y
 
 # Simulation
 numberlines = 200
 
 # Perform Line Sampling
-LS = LineSampling(numberlines, collect(0.5:0.5:10))
+LS = LineSampling(numberlines, collect(0.5:0.5:8))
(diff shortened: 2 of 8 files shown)


AnderGray (changes requested):
The changes look fine. However, if you uncomment the `RandomSlicing` part of the `demo/linesampling.jl`, it still errors. The error is however different this time, but indicates that there is still some instability. 

This demo is how we originally found this bug.

https://github.com/FriesischScott/UncertaintyQuantification.jl/blob/41a89ea33bfdfa7325136057a09bc8ada5bed474/demo/reliability/linesampling.jl#L66

src/simulations/linesampling.jl
```diff
@@ -8,6 +8,10 @@ mutable struct LineSampling <: AbstractSimulation
         points::Vector{<:Real}=collect(0.5:0.5:5),
         direction::NamedTuple=NamedTuple(),
     )
+
+    if maximum(points) > 8.0
+        error("LineSampling does not support lines longer than 8.12.")
```
AnderGray, line 13: Seems weird that the error says `8.12`, but the criteria is `8.0`. Maybe make them the same?

AnderGray (changes requested):
Other than the minor comment, I think ready to merge.

src/simulations/linesampling.jl
FriesischScott: Oops, my bad. That was not intended.

AnderGray:
Thanks for the fix, and the changes. Looks good!

PR #239, opened 2024-12-17
Add custom kernel density estimation

FriesischScott:
This gets rid of the **KernelDensity** dependency. The main annoyance with *KernelDensity* was the *FFTW* dependency which in turn pulled in *MKL*.

The method used to select the bandwidth for the KDE is taken from *S. J. Sheather and M. C. Jones, ‘A Reliable Data-Based Bandwidth Selection Method for Kernel Density Estimation’, Journal of the Royal Statistical Society: Series B (Methodological), vol. 53, no. 3, pp. 683–690, 1991, doi: [10.1111/j.2517-6161.1991.tb01857.x](https://doi.org/10.1111/j.2517-6161.1991.tb01857.x).*

Along with the PDF I also changed the CDF and inverse CDF.  Before we used the native Julia `ecdf` function for the CDF and interpolated the `quantile` from the available data and `ecdf` values, but I'm not a fan because it's not a smooth function. Now I've implemented numerical integration of the actual underlying PDF using `QuadKG` for the CDF and the inverse through root finding via *Root*. Two new dependencies, but they are pretty lightweight so I think its fine. The biggest benefit here is, that now the CDF and quantile are actual inverses so ` quantile(cdf(x) == x` which didn't really hold before.

  We'll have see if we need more efficient approaches in the future, but the accuracy is of the charts in comparison to the previous implementation. I've also added docs, docstrings and more stable tests.


docs/src/manual/kde.md
```diff
@@ -0,0 +1,43 @@
+# Kernel Density Estimation
+
+Kernel density estimation (KDE) is a non-parametric method to estimate the probability density function  of a random variable through *kernel smoothing* [silvermanDensityEstimationStatistics1986](@cite).
```
AnderGray, line 3: There's an extra space in

"probability density function __ of"

docs/src/manual/kde.md
```diff
@@ -0,0 +1,43 @@
...
+
+The kernel density estimate ``\hat{f}_h`` of a univariate density `f` based on a random sample ``X_1,\ldots,X_n`` is defined as
+
+```math
+\hat{f}_h(x) = n^{-1} \sum_{i=1}^n h^{-1} K \left\{\frac{x-X_i}{h}\right\},
+```
+
+where ``h`` is the so called *bandwidth* and ``K`` is the kernel function. The kernel function is assumed to be a symmetric probability density and is set to be a Gaussian density in *UncertaintyQuantification.jl*. The bandwidth ``h`` also called the *smoothing parameter* has a strong effect on the resulting density estimate. There are various different methods to select an optimal bandwith. Here we have decided to apply the method developed by Sheather & Jones [sheatherReliableDataBasedBandwidth1991](@cite) for its excellent performance and straightforward implementation.
+
+The kernel density estimation is exposed through the [`EmpiricalDistribution`](@ref). Since the banwidth is automatically selected only a vector containing the data must be passed to the constructor.
```
AnderGray, line 13: banwidth -> bandwidth

src/util/kde.jl
```diff
@@ -0,0 +1,53 @@
...
+    val = zero(eltype(x))
+
+    for xᵢ in x, xⱼ in x
+        val += ϕ4(α^(-1) .* (xᵢ .- xⱼ))
+    end
+
+    return (n * (n - 1))^(-1) * α^(-5) * val
+end
+
+function sheather_jones_bandwidth(x::AbstractVector)
```
AnderGray, line 33: Perhaps worth a quick mention of what the `sheather_jones_bandwidth` is

AnderGray (commented):
Great improvement!

The only concern is that the quantile function is very slow. Finding it takes ~23s to generate 100 samples.

`cdf` is also quite slow, but I guess `quantile` is more of a concern, because you're wrapping the `quadgk` integrator inside a root finder. This also affects `to_standard_normal_space!` and `to_physical_space!`

Not sure if I have a suggestion for a better way todo it though.

FriesischScott:
To increase the efficiency I've reworked it like this:

1. bandwidth `h` is obtained from the data
2. `lb` and `ub` are found from the `pdf` using root finding
3. the cdf is computed for a range of 10000 (default) values between `lb` and `ub` through numerical integration 
4. the `cdf` and `quantile` functions are linearly interpolated from the computed cdf values

This is incredibly fast and should be very close to the accuracy of the previous approach

AnderGray:
Thanks for the changes. Good to merge once the tests pass. Although the error looks unrelated to this PR

Issue #190, opened 2024-09-23
Gaussian Processes


FriesischScott:
Most likely based on AbstractGPs.jl

PR #245, opened 2025-02-11
Gaussian processes

felixmett:
**Related Issue**:
- This PR introduces a draft for the implementation of Gaussian processes, as outlined in issue #190.

**Description**:
- This PR defines a Gaussian process using `AbstractGPs.jl`, where you can specify a mean function, a kernel function, and optionally add observational noise.
- The Gaussian process can be fitted to data directly or have its hyperparameters optimized using maximum likelihood estimation (MLE).
- Hyperparameters are automatically extracted from the Gaussian process model during optimization.

**Current State**:
- A gaussian process can be fitted to data stored in a `DataFrame`.
- A gaussian process can be fitted with an `UQInput`, a `UQModel` and an experimental design.
- The hyperparameters of the Gaussian process can optionally be optimized using MLE.
- Both input and output standardization are supported.

**What’s Left to Do**:
- The automatic hyperparameter-extraction is not implemented for every possible mean and kernel function.
- The current optimization method is unstable and heavily relies on a good initial guess for the hyperparameters.
- Occasionally, `UQInput` standardization can break the fitting procedure due to `-Inf` or `Inf` values arising from the inverse standard normal CDF.
- There is no documentation yet
- There are no recent tests implemented


FriesischScott (changes requested):
On top of the comments please make the following changes:
- For the optimization use `ADTypes` and `AutoZygote()`.
- let's drop the output standardization for now.
- You might be able to use `supertypes` to get all kernel types and write a test to check that the necessary functions exist for all of them to extract and apply the parameters.

jgrashorn:
Things to do for Merge:
- [ ] Implement noise directly in GP
    - [ ] Add or un-add sigma in optimization [here](https://github.com/felixmett/UncertaintyQuantification.jl/blob/59822779def1e729ee43d86d8650208713125a1b/src/models/gp/parameterization.jl#L199) 
- [ ] Refine user interface
- [ ] Schlepp data with
- [ ] **_Add refinement_**
- [ ] Check and aufräum data transformation

FriesischScott:
Closed in favour of #334.

PR #248, opened 2025-05-28
Rework ProbabilityBox and Interval

### LINKED ISSUE #196: P-box and imprecise docs
We're missing theory and methodology docs for all of the double-loop and interval MC stuff

FriesischScott:
This PR introduces two main changes:

1. The `ProbabilityBox` is now without `name` and has to be wrapped inside a `RandomVariable` (now a parametric type)
 - Additionally, a pbox is now constructed using a Dictionary mapping the parameters of the distribution to either a `Real` or a `Interval`
 - I think this is a lot cleaner than before and also easier to handle in the code. 
3. The `Interval` type is now mainly used internally for the propagation of epistemic uncertainties and to construct p-boxes
 - A new `IntervalVariable`  is introduced to be used as inputs

This gives us more flexibility and allows us to use intervals and pboxes inside other structs in downstream packages.

As an added benefit this gives us dependent pboxes.

I've also added documentation on imprecise reliability analysis.

Closes #196

### DIFF
--- a/.github/workflows/DemoFilesUpToDate.yml
+++ b/.github/workflows/DemoFilesUpToDate.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
-          version: '1.10'
+          version: "1.11"
       - name: Install dependencies
         run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
       - name: Check demo files
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v2
         with:
-          version: "1.10"
+          version: "1.11"
       - name: Build and deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/DocumentationPreview.yml
+++ b/.github/workflows/DocumentationPreview.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v2
         with:
-          version: "1.10"
+          version: "1.11"
       - name: Build and deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         julia-version:
-          - "1.10"
+          - "1.11"
         os:
           - "ubuntu-latest"
           - "windows-latest"
@@ -29,7 +29,7 @@ jobs:
       - uses: julia-actions/cache@v1
       - name: Update PATH on Linux
         if: runner.os == 'Linux'
-        run: echo "$PWD/test/test_utilities" >> $GITHUB_PATH 
+        run: echo "$PWD/test/test_utilities" >> $GITHUB_PATH
       - uses: julia-actions/julia-runtest@latest
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v4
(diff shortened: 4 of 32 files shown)


AnderGray (commented):
Very nice to have dependent p-boxes working!

Overall I think it's nice. I do find it slightly confusing that we now need to wrap the p-box type in a `RandomVariable`. Also at some point people will probably be confused by the distinction between `Interval` and `IntervalVariable`.

But I guess the distinction is required for other codes to use this features?

* I also notice that the `mean` and `var` functions for p-boxes and intervals return errors:

``
s = ProbabilityBox{Normal}(Dict(:μ => Interval(-1, 1), :σ => Interval(1, 2)))
ProbabilityBox{Normal}(Dict{Symbol, Union{Real, Interval}}(:μ => [-1, 1], :σ => [1, 2]), -Inf, Inf)

mean(s)
var(s)
``
Similarly for the `interval`, `IntervalVariable`, and `RandomVariable(ProbabilityBox{Normal}(Dict(:μ => Interval(-1, 1), :σ => Interval(1, 2)))`

* The demo file `demo/linesampling.jl` needs to be updated to reflect the new constructors

FriesischScott:
> Very nice to have dependent p-boxes working!
> 
> Overall I think it's nice. I do find it slightly confusing that we now need to wrap the p-box type in a `RandomVariable`. Also at some point people will probably be confused by the distinction between `Interval` and `IntervalVariable`.
> 
> But I guess the distinction is required for other codes to use this features?

Yes this allows not only other but also us to use p-boxes and intervals internally where we might not necessarily have an associated `Symbol`. 

I'll work on the documentations to make the distinction clear to the user.

AnderGray (changes requested):
Thanks for the changes.

Finding the following two edge cases:

```
par = Dict(:a => Interval(0.14, 0.16), :b => 0.16)
p_box = ProbabilityBox{Uniform}(par)

mean(p_box)
```

errors: `ERROR: ArgumentError: Uniform: the condition a < b is not satisfied.`

```
a = IntervalVariable(0, 1, :a)

var(a)
```
the `var` for an interval hasn't been defined. I think it's the interval `[0, (hi-lo)^2 /4]`

Otherwise good to go for me

AnderGray (approved):

Rucksack94:
I see the benefit of the types changes, but it prohibits the usage of different univariate distributions in TMCMC due to the prior type change from `prior::Vector{RandomVariable}` to `prior::Vector{<:RandomVariable{<:UnivariateDistribution}}` due to Julia types.

**MWE:**
```
julia> v1 = RandomVariable.([Normal(), Normal()],[:x1,:x2])
julia> v2 = RandomVariable.([Uniform(), Uniform()],[:x1,:x2])
julia> v3 = RandomVariable.([Normal(), Uniform()],[:x1,:x2])

julia> typeof(v1)
Vector{RandomVariable{Normal{Float64}}}
julia> typeof(v2)
Vector{RandomVariable{Uniform{Float64}}}
julia> typeof(v3)
Vector{RandomVariable}

julia> isa.([v1,v2,v3],Vector{<:RandomVariable{<:UnivariateDistribution}})
3-element BitVector:
 1
 1
 0
```
I don't_ know if that was intended or I missed something else, just broke my code updating to UQ.jl v0.14.0.

FriesischScott:
> I see the benefit of the types changes, but it prohibits the usage of different univariate distributions in TMCMC due to the prior type change from `prior::Vector{RandomVariable}` to `prior::Vector{<:RandomVariable{<:UnivariateDistribution}}` due to Julia types.
> 
> **MWE:**
> ```
> julia> v1 = RandomVariable.([Normal(), Normal()],[:x1,:x2])
> julia> v2 = RandomVariable.([Uniform(), Uniform()],[:x1,:x2])
> julia> v3 = RandomVariable.([Normal(), Uniform()],[:x1,:x2])
> 
> julia> typeof(v1)
> Vector{RandomVariable{Normal{Float64}}}
> julia> typeof(v2)
> Vector{RandomVariable{Uniform{Float64}}}
> julia> typeof(v3)
> Vector{RandomVariable}
> 
> julia> isa.([v1,v2,v3],Vector{<:RandomVariable{<:UnivariateDistribution}})
> 3-element BitVector:
>  1
>  1
>  0
> ```
> I don't_ know if that was intended or I missed something else, just broke my code updating to UQ.jl v0.14.0. 
> 
> 

This should already be fixed on `master` and the fix will be included in the next release. We'll likely tag `0.15.0` in a few days.

PR #252, opened 2025-06-24
Bmu laplace estimate

jgrashorn:
Added a small addition to the point estimates. From a maximum a posteriori approximation the Hessian can be calculated and its inverse is then the covariance matrix of some Gaussian approximation. Also works on multimodal posterior distributions by using multiple starting points for the optimization. The Gaussian components are then weighted according to their relative posterior values.

Currently the output is a `DataFrame` that holds the MAP estimates and the covariance matrices, as well as a second output with a function handle for the (log-)posterior. Can be enhanced by using @lukasfritsch implementation of Gaussian mixtures. Also would benefit from having some better approximation of the Hessian. I used a finite differences-approximation currently that was adopted from old Matlab-code. But this surely can be enhanced with automatic differentiation or based on `FiniteDiff.jl` or `FiniteDifferences.jl`. Then only this line needs to be changed:

```julia
hess = [inv(fd_hessian(optimTarget, var, fddist)) for var in eachrow(vars)]
```

Also lacks documentation right now, but it works very nicely on some examples.


FriesischScott:
We're most likely going to use `Zygote` for AD with the GPs. Should be a good candidate for the hessian here as well.

jgrashorn:
It does, yes.
I was waiting for the `JointDistribution` to be merged so that I can use it.

FriesischScott (changes requested):
Still needs to be updated to return a `JointDistribution` containing a `MixtureModel` like the Gaussian Mixture Model does.

It also needs tests and a bit of documentation.

FriesischScott (changes requested):

FriesischScott (approved):

PR #253, opened 2025-06-30
Create design of experiments from quantiles

FriesischScott:
Instead of scaling by a factor of standard deviations the design is scaled to 99% quantiles (default). This way we don't need to treat bounded/unbounded distributions differently.

This should safeguard against the  `Inf`/`-Inf` values in standard normal space we encountered before. 

Does the user even need the ability to change the quantiles? We could increase the value a bit more and keep it fixed.

### DIFF
--- a/src/simulations/doe.jl
+++ b/src/simulations/doe.jl
@@ -1,78 +1,64 @@
 
 struct TwoLevelFactorial <: AbstractDesignOfExperiments
-    σ::Int
-    function TwoLevelFactorial(σ::Int=1)
-        σ < 1 && error("sigma must be >= 1")
-        return new(σ)
+    q::Real
+    function TwoLevelFactorial(q::Real=0.99)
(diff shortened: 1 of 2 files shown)


src/simulations/doe.jl
```diff
@@ -61,41 +61,25 @@ function full_factorial_matrix(levels::Vector{<:Integer})
...
-    samples = doe_samples(design, n_rv)
-    b = reduce(vcat, bounds.(random_inputs, design.σ))
-
-    for i in 1:n_rv
-        samples[:, i] = samples[:, i] .* (b[i, 2] - b[i, 1]) .+ b[i, 1]
-    end
-
+    # samples are transformed to be in [1-q,q]
+    samples = max.(doe_samples(design, n_rv) .* design.q, 1.0 - design.q)
+    samples = quantile.(Normal(), samples)
```
teobros, line 73: The idea to go to SNS and then to physical to cover all the possible complex cases of dependencies and imprecision is indeed very clever!

Issue #209, opened 2024-09-25
Naming convention of random variables in spectral representation


mbvoyager:
Samples of stochastic processes generated using spectral representation are now all labeled by the symbol `:gm`.

This results in the naming of the collection of generated signals (stochastic processes) as `gm` and the random phase angles as `gm_1, gm_2, gm_3, ...`.
This is confusing. Since at least for spectral representation only the phase angles `ϕ` are the random variables, I think it would be better to name and number them as `ϕ_1, ϕ_2, ϕ_3, ...`

![image](https://github.com/user-attachments/assets/cbcd55f8-5590-4846-9048-d301a8f2c732)

FriesischScott:
Since we need a reference to the name of the spectral representation `gm_ϕ_1` would be the obvious choice.

I have a few more ideas on how we can potentially improve the PSD and SRE code:
- Don't discretize the PSD but keep it as a proper function that could be called as `PSD(ω)`
  - This could a candidate for memoization so we can cache repeated calls for the same `ω`
- Instead of setting `ω` as a `Vector` we could define the cutoff frequency and number of discretization points `N` in the  `SpectralRepresentation`  and then get the vector from that whenever needed. Might be cleanier for the struct.


A question about `N`. You need a relatively large `N` for the stochastic process to be Gaussian. In one of the tests we're using `100` which is insufficient. Is there any recommendation or rule of thumb for how large `N` should be?

Thoughts @mbvoyager, @marcobehrendt?

PR #260, opened 2025-08-08
Add Nyquist frequency violation warning, and change naming of RVs in SRM method

mbvoyager:
Hi,

Mira and I implemented a Nyquist frequency violation warning inside the generation of the `spectralrepresentation` object. To warn the user that the time and frequency discretization may cause aliasing.

Additionally, Mira added the `varphi` to the naming of the RVs inside the `spectralrepresentation` method.

Here is how it looks like now:
image

We also added a test that uses `test_logs` to check if the warning correctly appears.

These implementations and changes are referring to the Issue https://github.com/FriesischScott/UncertaintyQuantification.jl/issues/209 .


FriesischScott (commented):
Great stuff 👍 . I think it would be great  if you could add  some info on the Nyquist frequency  and what it means to the documentation.

FriesischScott (approved):

Issue #82, opened 2022-08-26
Sliced distributions


FriesischScott:
Separate package but wrapped in here.

FriesischScott:
With the package finally registered we can bring these in.

FriesischScott:
With the `JointProbability` now accepting any `MultivariateDistribution` (#261) sliced distributions work out of the box.

PR #261, opened 2025-08-11
Rework JointDistribution

### LINKED ISSUE #226: Interface for joint distributions
We need a general interface for various joint distributions. I think a parameterized struct could work well here. In this we can wrap for example `SklarDist` from Copulas.jl, `MvNormal` from `Distributions.jl` and Sliced Normals/Exponentials.

```julia
struct JointDistribution{T} <:RandomUQInput begin
	dist::T
	names::Vector{Symbol}
```

This would allow us to define mappings for dispatch for distinct types.

```julia
function to_standard_normal_space!(JointDistribution{MvNormal}, df::DataFrame)
...
end

function to_standard_normal_space!(JointDistribution{SlicedNormal}, df::DataFrame)
	error("SlicedNormals can not be mapped to standard normal space.")
end
```

FriesischScott:
The main change in this PR is making the `JointDistribution` parametric. This allows for two distinct types of joint distributions defined from either a `Copula` and a `Vector{RandomVariable}` as before or  (new) a `MultivariateDistribution` and a `Vector{Symbol}`. 

The latter provides the possibility to wrap all sorts of multivariate distributions such as the Gaussian mixture models provided in #257 and sliced distributions, see #82.

closes #226.

I've added the appropriate tests and documentation.

### DIFF
--- a/test/reliability/probabilityoffailure_imprecise.jl
+++ b/test/reliability/probabilityoffailure_imprecise.jl
@@ -28,16 +28,31 @@
                 :Y,
             )
 
-            c = GaussianCopula([1 0.0; 0.0 1])
+            @testset "independent" begin
+                c = GaussianCopula([1 0.0; 0.0 1])
 
-            jd = JointDistribution(c, [X, Y])
+                jd = JointDistribution(c, [X, Y])
 
-            pf, x_lb, x_ub = probability_of_failure(
-                UQModel[], df -> 9 .+ df.X .+ df.Y, jd, DoubleLoop(FORM())
-            )
+                pf, x_lb, x_ub = probability_of_failure(
+                    UQModel[], df -> 9 .+ df.X .+ df.Y, jd, DoubleLoop(FORM())
+                )
 
-            @test pf.lb ≈ cdf(Normal(2, sqrt(2)), -9) atol = 1e-6
-            @test pf.ub ≈ cdf(Normal(-2, sqrt(8)), -9) atol = 1e-6
+                @test pf.lb ≈ cdf(Normal(2, sqrt(2)), -9) atol = 1e-6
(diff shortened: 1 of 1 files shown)


src/inputs/jointdistribution.jl
```diff
@@ -1,60 +1,93 @@
-struct JointDistribution <: RandomUQInput
-    marginals::Vector{<:RandomVariable}
-    copula::Copula
-
-    function JointDistribution(marginals::Vector{<:RandomVariable}, copula::Copula)
-        length(marginals) == dimensions(copula) ||
-            error("Dimension mismatch between copula and marginals")
-
-        return new(marginals, copula)
+struct JointDistribution{
```
AnderGray, line 36: This type definition might be slightly more readable: 

```julia
struct JointDistribution{D<:Union{Copula,MultivariateDistribution}, M} <: RandomUQInput
    d::D
    m::Vector{M}

    # Copula + RandomVariable
    function JointDistribution(d::C, m::Vector{RV}) where {C<:Copula, RV<:RandomVariable}
        dimensions(d) == length(m) ||
            throw(ArgumentError("Dimension mismatch between copula and marginals."))
        new{C,RV}(d, m)
    end

    # MultivariateDistribution + Symbol
    function JointDistribution(d::MD, m::Vector{Symbol}) where {MD<:MultivariateDistribution}
        length(d) == length(m) ||
            throw(ArgumentError("Dimension mismatch between distribution and names."))
        new{MD,Symbol}(d, m)
    end
end
```

AnderGray: The following test may also no longer be needed:

```julia
@test_throws ErrorException("Must pass a vector of Symbols.") JointDistribution(
dist, [RandomVariable(Normal(), :x), RandomVariable(Uniform(), :y)])
```

AnderGray (approved):

PR #276, opened 2025-10-16
Feature/kucherenko indices

sitoryu:
Implementation of 2 methods to calculate Kucherenko Indices and ability to create conditional samples from a gaussian copula

New functions:

Standard Kucherenko Indices using conditional sampling
Sample based Kucherenko Indices using binning
create conditional samples from a gaussian copula


src/sensitivity/kucherenkoindices.jl
```diff
@@ -0,0 +1,394 @@
...
+    
+    return ST_i
+end
+
+function _generate_conditional_samples(
+    samples::DataFrame,
+    joint_dist::JointDistribution,
+    var_names::Vector{Symbol}
+)
+    input_var_names = [marginal.name for marginal in joint_dist.m]
```
FriesischScott: This is just `names(joint_dist)`.

src/sensitivity/kucherenkoindices.jl
```diff
@@ -0,0 +1,394 @@
...
+
+"""
+- `models::Vector{<:UQModel}`: Vector of UQ models to evaluate
+- `inputs::Vector{<:UQInput}`: Vector of input distributions
+- `outputs::Vector{Symbol}`: Vector of output quantity names
+- `sim::AbstractMonteCarlo`: Total MC samples
+- `min_bin_sample::Int=25`: Minimum samples per bin for 1 dimension for conditioning; Recommended amount is at least 25 samples per bin but bin amount not higher than 100
+- `min_bin_sample_multi_dims::Int=25`: Minimum samples per bin-dimension in multiple dimensions for conditioning; Recommended amount is about 10-25 samples per bin
+"""
+function kucherenkoindices_bin(
```
FriesischScott: By changing the first method to a `DataFrame` you can get rid of all of these and it will be up to the user if they want to use the conditional sampling or the binning.

FriesischScott (changes requested):
Please move both the conditional sampling and the corresponding tests into  joint distribution files.

Overall we need to be able to pass both dependent and independent inputs to the `kucherenkoindices` function at the same time so we can for example compute the sensitivity for the cantilever example.

PR #279, opened 2025-10-23
Add basic plots

### LINKED ISSUE #191: Add plot recipes
Add some plotting recipes for, e.g.:

* Inputs (random variables, inputs, p-boxes, bivariate distributions).
* A sampled dataframe of input / outputs.
* Some simulation methods (e.g., various levels of subset simulations).
* Others?

AnderGray:
This PR adds the ability plot the following using plot recipes: 

* Distributions (both pdf and cdf)
* IntervalVariables
* P-boxes
* A vector of UQInputs (plots it in a grid)
* A vector of Intervals (the ecdf of a p-box)
* A bivariate scatter of Intervals (ecdf of a bivariate p-box)

With plot recipes, various properties of the plots should be editable by the user as they would any standard plot with Plots.jl (i.e. `color=:red`, will make the plot red)

Eg, the following DEMO script (in `demo/plotting`):

```julia
using UncertaintyQuantification, Plots

X1 = RandomVariable(ProbabilityBox{Normal}(Dict(:μ => Interval(-1, 2), :σ => 1)), :X1)
X2 = RandomVariable(ProbabilityBox{Normal}(Dict(:μ => Interval(-1, 2), :σ => 2)), :X2)
X3 = IntervalVariable(-1, 2, :s)
h = RandomVariable(Normal(0.24, 0.01), :h) # height

plot(X1)    # p-box
plot(X1, color = "red") # red p-box
plot(X3)    # Interval
plot(h)     # distribution 

inputs = [X1, X2, X3, h]

plot(inputs)    # Everything together

samples = sample(inputs, 50)

plot(X1)
plot!(samples.X1)    # Interval ecdf of X1

plot(samples.X1[1], samples.X2[1])  # Plots 2D box

plot(samples.X1, samples.X2)        # Plots bivariate random sets of X1 and X2

```

plot_test1
plot_test_2
plot_test_3


FriesischScott (changes requested):
Overall very nice plots!

I do have some things I'd prefer differently though:

 - by default I think the colour scheme should be the same as when plotting regularly, i.e. cycle through the default colours  (blue, orange,...) with each extra plot added to a single figure. This way the colours are consistent with any other stuff the user might plot.
-  I don't really like the grey shaded area in all plots. For a PDF for example just the line should be sufficient. Maybe for the pbox and interval it makes sense though.
-  Instead of the `xlabel` I think the variable name should actually go in the plot label. Currently if I plot two pdfs in one figure the second one will override the name of the first. Without passing colours, they will also be the same colour.

Please also add a few plots in suitable areas of the documentation. For example where the inputs are introduced.

FriesischScott (changes requested):
Very nice!Almost ready to merge, but  I found two minor issues I'd like to resolve before merging:

- The empirical cdf plot from interval samples still has very thin lines and no shaded area like the pbox plots. Could be good to bring the closer together visually.
-  In the random slicing demo file you pass `label` to a few plots but it's ignored and there is no legend in the final plots.

FriesischScott (approved):

PR #281, opened 2025-11-16
Generalize pcebases and multiple PCE outputs

fbelik:
This is based on discussion in Issue [251](https://github.com/JuliaUQ/UncertaintyQuantification.jl/issues/251).

1. Speed up `evaluate` call for a `PolynomialChaosBasis`.
2. Implement various multi-index sets, now can pass a symbol to `PolynomialChaosBasis` for the user to specify a total-degree set (was and is still default, ordering just changes), total product set, hyperbolic cross set (sparse alternative), or a "q-ball set" (the Euclidean q-norm of the indices is less than `p`). 
3. Decluttering of `polynomialchaos` methods; previously several methods for different types of inputs but were able to be collapsed down into a single method per approach (LS,GQ,etc).
4. Can pass a vector of outputs to `polynomialchaos` to generate a vector of `PolynomialChaosExpansion` objects. Potentially useful when several models are used on the same type of input with different output. Does not break current code as if only a single output is passed, a single `PolynomialChaosExpansion` object is returned.


FriesischScott:
Thanks for another PR @fbelik! I wasn't able to review it yet, but I'll definitely get to it this week.

src/models/pce/polynomialchaosexpansion.jl
```diff
@@ -29,12 +29,15 @@ end
...
-    inputs::Vector{<:UQInput},
-    model::Vector{<:UQModel},
+    inputs::Union{<:UQInput,Vector{<:UQInput}},
+    model::Union{<:UQModel,Vector{<:UQModel}},
     Ψ::PolynomialChaosBasis,
-    output::Symbol,
+    outputs::Union{Symbol,<:AbstractVector{Symbol}},
     ls::LeastSquares,
 )
+    inputs = isa(inputs, UQInput) ? [inputs] : inputs
```
FriesischScott: We have utility function `wrap` for this. It returns vectors or wraps scalars into a vector.

FriesischScott (changes requested):
Overall very good work 👍 . Just a few comments.

Please add tests for the q-ball set.

Do you have suggested literature for someone not familiar with these alternative index sets? I'd like to know a bit about what they are used for.

src/models/pce/pcebases.jl
```diff
@@ -6,14 +6,26 @@ struct PolynomialChaosBasis
...
 
-    function PolynomialChaosBasis(bases::Vector{<:AbstractOrthogonalBasis}, p::Int)
+    function PolynomialChaosBasis(bases::Vector{<:AbstractOrthogonalBasis}, p::Int, index_set::Symbol=:TD; param=0.5)
         d = length(bases)
-        return new(bases, p, d, multivariate_indices(p, d))
+        return new(bases, p, d, multivariate_indices(p, d, index_set; param=param))
+    end
+
+    # in_index_set must take in (idx,p) and return boolean whether or not in multi-index set
+    function PolynomialChaosBasis(bases::Vector{<:AbstractOrthogonalBasis}, p::Int, in_index_set::Function)
```
FriesischScott: What is the purpose of this method? It doesn't look like it's being used right now.

src/models/pce/pcebases.jl
```diff
@@ -59,30 +71,67 @@ function He(x::Real, n::Integer)
...
-    end
-
-    P = d + 1
-    for k in 2:p
-        L = P
-        for j in 1:d, m in (L - pᵢ[j, k] + 1):L
-            P += 1
-            idx[P, :] = idx[m, :]
-            idx[P, j] = idx[P, j] + 1
+        if all(iszero, idx)
```
FriesischScott: Not much of a difference but `iszero` accepts arrays and returns `true` if all elements are zero. So `iszero(idx)` works here.

src/models/pce/pcebases.jl
fbelik: We can get rid of this one if you prefer. It allows the user to pass in their own `in_index_set` method to create PCE bases for arbitrary multi-index sets which may not be implemented in the package. The goal was to provide the user some additional flexibility in case a user wishes to use a multi-index set not implemented in the package. An alternative route would be to make the inner constructor `PolynomialChaosBasis(bases::Vector{<:AbstractOrthogonalBasis}, p::Int, index_set::Symbol=:TD; param=0.5)` an outer constructor.

src/models/pce/pcebases.jl
```diff
@@ -59,30 +71,67 @@ function He(x::Real, n::Integer)
...
+end
 
-    for k in 2:No
+# Q-ball multi-index set (Sparse alternative to TD or TP when q < 1, same as TD when q=1, same as TP when q=Inf)
+function QB(idx::Vector{Int}, p::Int, q::Real=0.5)
+    @assert q > 0
+    return norm(idx, q) <= p
+end
+
+function multivariate_indices(p::Int, d::Int, in_index_set::Function=TD; max_size=1_000_000_000_000)
```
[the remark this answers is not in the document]
fbelik: This could probably be replaced by the size of the total product set, `p^d`, which I would maybe write as `BigInt(p)^d` to prevent integer overflow. That should be the largest, downward-closed multi-index set for fixed `p` and `d`.

Issue #271, opened 2025-10-03
IntervalPredictorModel


FriesischScott:
### References

[1] L. G. Crespo, S. P. Kenny, and D. P. Giesy, ‘Interval Predictor Models With a Linear Parameter Dependency’, Journal of Verification, Validation and Uncertainty Quantification, vol. 1, no. 021007, Jan. 2016, doi: [10.1115/1.4032070](https://doi.org/10.1115/1.4032070).

PR #305, opened 2026-04-10
LinearBasisFunctionModels

### LINKED ISSUE #272: Redesign ResponseSurface
The `ResponseSurface` and `PolyharmonicSpline`	essentially do the exact same thing with  different basis. I suggest we refactor them into a single type that is passed the desired basis. This can include other bases like a Gaussian radial basis.

We can then use these in the `IntervalPredictorModel`.

FriesischScott:
This PR introduces the `LinearBasisFunctionModel` as a generalization of the `ResponseSurface` and similar models.

While the `PolyharmonicSpline` is technically a radial basis with an added polynomial term it requires a separate type because of added constraints. For consistency I have created an alias `ResponseSurface` for the linear basis function model with a monomial basis.

Overall this will make it easier to introduce similar models such as radial basis function networks and especially interval predictor models (#271).

As of now, this includes no new functionality and the existing models should function exactly as before.


docs/literate/metamodels/responsesurface.jl
```diff
@@ -70,7 +70,8 @@ That functions degree is set as an Integer in the constructor.
 
 training_data = sample(x, design)
 evaluate!(himmelblau, training_data)
-rs = ResponseSurface(training_data, :y, 4)
```
AnderGray, line 73: For the tutorial, do you think it is worth keeping `ResponseSurface` in the original form you used?

The newer formulation is more general, but it also requires a few more lines. Perhaps we could keep the simpler `ResponseSurface` version in the tutorial, and mention that a `ResponseSurface` is essentially a `LinearBasisFunctionModel` with a `MonomialBasis`.

In any case, the text just below this section still refers to it as a `ResponseSurface`, so we may want to make the terminology consistent.

AnderGray (approved):

PR #308, opened 2026-05-04
JointDensity as return value for Laplace estimate

jgrashorn:
As suggested by @lukasfritsch it would be better if the Laplace estimate actually returned a JointDensity, so further evaluation with the package is easily possible without having to manually wrap it.
The PR changes this and also fixes the tests and documentation for the change.


FriesischScott (approved):

FriesischScott:
Perfect, thanks!

PR #332, opened 2026-07-27
Restore isimprecise for vector of inputs

### LINKED ISSUE #330: ```isimprecise``` not working on vector of ```UQInput```
With version 0.15 ```isimprecise``` function does not dispatch over ```Vector{UQInput}```. 
The new method requires a ```UQModel``` too, but would nice and painless to still dispatch over just ```UQInputs```

FriesischScott:
I'm also exporting the function and added docstrings.

Closes #330.


teobros (commented):
I think it is good.

teobros (approved):

